PR #17, opened 2015-11-03
add Shallow query

famersbs:
if you use shallow options, ignore every query


jeromegamez:
Thank you for your contribution @famersbs - and sorry that it took me so long to reply - this PR must somehow have slipped through :/.

I have used your initial implementation (without the tabs to spaces changes and with little refactoring + tests) in #18 and will make a new release as soon as Scrutinizer and Travis are okay with what I did there.

Thanks again, and please find yourself in the changelog :bouquet:

PR #23, opened 2016-06-27
Small PHP doc fix : limit to last x items

FrancoisMartin:
Hello,

I think the review will be pretty fast :)
Thanks for this great lib.


jeromegamez:
Merci François!

PR #63, opened 2017-02-17
Update Reference.php: check if reference has child

spam312sn:
Checking if exist child of reference

### DIFF
--- a/src/Firebase/Database/Reference.php
+++ b/src/Firebase/Database/Reference.php
@@ -252,6 +252,22 @@ class Reference
     }
 
     /**
+     * Returns if exist children of a reference.
+     * 
+     * @param string $child
+     *
+     * @throws ApiException if the API reported an error
+     *
+     * @return boolean
+     */
+    public function hasChild($child = ''): boolean
+    {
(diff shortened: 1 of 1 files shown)


jeromegamez:
Thanks for your contribution, @spam312sn!

A `hasChild()` method already exists in the [`Firebase\Database\Snapshot` class](https://github.com/kreait/firebase-php/blob/master/src/Firebase/Database/Snapshot.php#L115).

Fetching the snapshot first and then checking has the advantage that you don't have to perform actual HTTP requests each time you want to check for a child key:

```php
$checks = ['first', 'second', 'third'];
$ref = $database->getReference('my-ref');
$snapshot = $ref->getSnapshot();

foreach ($checks as $check) {
    // would execute one extra HTTP request in each loop 🙁
    $ref->hasChild($check); 
    // works on the already fetched Snapshot -> no extra requests 👍
    $snapshot->hasChild($check); 
}
```

Although I will not include this proposed change, I want to thank you for using the library and putting effort into giving back to the project!

spam312sn:
Sorry, I haven't noticed this

PR #64, opened 2017-02-17
Update ReferenceTest.php: new test for `hasChild()` method

spam312sn:
Second part of pull request #63
Test for new function `hasChild()`


jeromegamez:
Please see https://github.com/kreait/firebase-php/pull/63#issuecomment-280704343 for the reason why this PR got closed.

In general, it is considered a best practice to include a feature and its test(s) in the same pull request - it makes it easier to review all the changes that belong to each other and avoids failed builds like in this case, where the test references a method that is not available in this branch.

Thanks!

spam312sn:
Well, I'll remember for the future. Just created through GitHub and he has created a commit with only one file

PR #69, opened 2017-03-14
PHP Variable not echoing because it's not defined

johnparkerg:
changed $token for $customToken


jeromegamez:
Fixed with 07f1d8108a24e43e5ba0305fb74aee80559f5688 - thank you @johnparkerg!

PR #83, opened 2017-04-23
Firebase event stream

saishsali:
Streaming from the REST API: https://firebase.google.com/docs/reference/rest/database/#section-streaming


jeromegamez:
Hey @saishsali, is this still work in progress? No pressure, just curious :)

saishsali:
@jeromegamez : Yes. Will complete it by this weekend.

saishsali:
@jeromegamez : Can you please review the PR?

jeromegamez:
Hey @saishsali and thank you very much for your contribution! I just tested the functionality with the following script and it works just fine! 🎉

```php
require __DIR__.'/vendor/autoload.php';

use Kreait\Firebase\Database\ServerSentEvent;
use Kreait\Firebase\Factory;

$firebase = (new Factory())
    ->create()
    ->withStreaming();

/** @var ServerSentEvent[] $events */
$events = $firebase->getDatabase()->getReference('/')->getEvents();

$isFirstEvent = true;
foreach ($events as $event) {
    if ($isFirstEvent) {
        $isFirstEvent = false;
        continue;
    }

    printf("%s: %s\n", $event->getEventType(), $event->getData());
}
```

I will look at the code in depth soon, but this is definitely a great start! Thank you!

saishsali:
@jeromegamez Thanks a lot for testing the functionality. I will update the documentation once you are done reviewing! :)

src/Firebase/Database/SseClient.php
```diff
@@ -0,0 +1,109 @@
...
+            }
+        }
+    }
+
+    /**
+     * Connect to server
+     */
+    private function connect()
+    {
+        $headers = [];
```
jeromegamez: If you set the default headers here instead of in the `Kreait\Firebase` class 

```php
$headers = [
    'Accept' => 'text/event-stream',
    'Cache-Control' => 'no-cache',
];
```

you can revert the customizations in `src/Firebase.php` concerning the streaming. If works the same as before, but the developer doesn't need to differ between streaming and a non-streaming Firebase instance.

src/Firebase/Database/ApiClient.php
```diff
@@ -46,6 +46,11 @@ public function get($uri)
         return JSON::decode((string) $response->getBody(), true);
     }
 
+    public function stream($uri, $options)
```
jeromegamez:
```php
public function stream($uri, $options): ResponseInterface
```

src/Firebase/Database/ServerSentEvent.php
```diff
@@ -0,0 +1,132 @@
...
+        $this->data = $data;
+        $this->eventType = $eventType;
+        $this->id = $id;
+        $this->retry = $retry;
+    }
+
+    /**
+     * @param string $raw
+     *
+     * @return Event
```
jeromegamez: `ServerSentEvent`

src/Firebase/Database/ServerSentEvent.php
```diff
@@ -0,0 +1,132 @@
...
+            $matched = preg_match('/(?P<name>[^:]*):?( ?(?P<value>.*))?/', $line, $matches);
+
+            if (!$matched) {
+                throw new InvalidArgumentException(sprintf('Invalid line %s', $line));
+            }
+
+            $name = $matches['name'];
+            $value = $matches['value'];
+
+            if ($name === '') {
```
jeromegamez: This case is also handled by the `switch()` statement below when none of the defined cases applies.

src/Firebase/Database/ServerSentEvent.php
```diff
@@ -0,0 +1,132 @@
...
+                case 'data':
+                    $event->data = empty($event->data) ? $value : "$event->data\n$value";
+                    break;
+                case 'id':
+                    $event->id = $value;
+                    break;
+                case 'retry':
+                    $event->retry = (int) $value;
+                    break;
+                default:
```
jeromegamez: The empty `default` can be omitted as nothing happens.

Also, in PHP, `continue` inside a `switch` behaves like a `break` - to break out of the loop, we would have to use `continue 2;`. But to be honest, I have never seen come something good out of a `continue <x>` or `break <y>` 😅

src/Firebase/Database/SseClient.php
```diff
@@ -0,0 +1,109 @@
...
+        $this->apiClient = $apiClient;
+        $this->url = $url;
+
+        $this->connect();
+    }
+
+    /**
+     * Returns generator that yields new event when it's available on stream.
+     *
+     * @return Event[]
```
jeromegamez: `@return \Generator|ServerSentEvent[]` - the first is the actual return value, the second helps us developers to loop around ^^

src/Firebase/Database/SseClient.php
```diff
@@ -0,0 +1,109 @@
...
+    const END_OF_MESSAGE = "/\r\n\r\n|\n\n|\r\r/";
+
+    /**
+     * @var Client
+     */
+    private $apiClient;
+
+    /**
+     *
+     * @var GuzzleHttp\Psr7\Response
```
jeromegamez: I personally would prefer `Psr\Http\Message\ResponseInterface` here. Guzzle does return a `GuzzleHttp\Psr7\Response`, but if one day we change the ApiClient implementation, this saves us one line of PHPDoc that we have to change ^^.

src/Firebase/Database/SseClient.php
```diff
@@ -0,0 +1,109 @@
...
+    /**
+     * @param ApiClient $apiClient
+     * @param string    $url
+     */
+    public function __construct(ApiClient $apiClient, string $url)
+    {
+        $this->apiClient = $apiClient;
+        $this->url = $url;
+
+        $this->connect();
```
jeromegamez: This will start streaming as soon as the object is created (which could lead to unexpected behaviour) - moving it to the `getEvents()` method should make it work so that the connection is only done at the moment it's needed, perhaps like this?

```php
if (!$this->response) {
    $this->connect();
}
```

saishsali:
@jeromegamez : I have done the d

hgouveia:
Any news about this PR? i'm waiting for this functionality 👍

saishsali:
@hgouveia : Waiting for @jeromegamez to review it once!

PR #94, opened 2017-06-09
Use non deprecated method

sota1235:
Hi, I have one point to fix.

`Google_Client::setAuthConfigFile` is deprecated now. So I replace this method.
https://github.com/google/google-api-php-client/blob/master/src/Google/Client.php#L830

Actually I want to use latest version of your library, but my PHP version is not 7~.
So please consider about this PR.

Thank you :)

### DIFF
--- a/src/Configuration.php
+++ b/src/Configuration.php
@@ -140,7 +140,7 @@ class Configuration implements ConfigurationInterface
         }
 
         $client = new \Google_Client();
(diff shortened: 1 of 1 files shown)


jeromegamez:
Hi Sota, and thank you for your contribution! I just published a new  [Release 1.2.2](https://github.com/kreait/firebase-php/releases/tag/1.2.2) including your change!

:octocat:

sota1235:
Thank you so much!!

PR #105, opened 2017-07-20
Add Authentication and User Management features

jeromegamez:
This is a pull request to track the progress of implementing Firebase Authentication and User Management features (https://firebase.google.com/docs/reference/rest/auth/)

## How to test

⚠️ **Method names and signatures might change** ⚠️ 

If you want to test it, you can do this by adding/replacing this line in your `composer.json`:

```
"require": {
    ...
    "kreait/firebase-php": "dev-feature/auth"
    ...
}
```

You can then activate and use the Authentication sub API by using your project's Web API key with the Firebase factory:

```php
require __DIR__.'/vendor/autoload.php';

use Kreait\Firebase\Exception\AuthException;
use Kreait\Firebase\Factory;
use Kreait\Firebase\ServiceAccount;
use Kreait\Firebase\Auth\User;

$apiKey = '<api key>';
$serviceAccount = ServiceAccount::discover();

$firebase = (new Factory())
    ->withServiceAccountAndApiKey($serviceAccount, $apiKey)
    ->create();

$auth = $firebase->getAuth();

try {
    // passwords must be at least six characters long
    $user = $auth->createUserWithEmailAndPassword('user@domain.tld', 'foobar');
    $updatedUser = $auth->changeUserPassword($user, 'new password');
    
    $auth->deleteUser($updatedUser);
} catch (AuthException $e) {
    echo $e->getMessage().PHP_EOL;
}
```

As a rule of thumb, use an IDE with autocompletion and let it guide you :)

## Progress

**`Firebase\Auth::`**
- [x] `getUser($uid): User` (if it doesn't exist, a new one will be created)
- [x] `createUserWithEmailAndPassword($email, $password): User`
- [x] `getUserByEmailAndPassword($email, $password): User`
- [x] `createAnonymousUser(): User`
- [x] `changeUserPassword(User $user, string $newPassword): User` 
- [x] `changeUserEmail(User $user, string $newEmail): User`
- [x] `deleteUser(User $user): void`
- [x] `sendEmailVerification(User $user): void`
- [x] Update documentation

:octocat:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # CHANGELOG
 
+## Unreleased
+
+* Added user management features ([Documentation](https://firebase-php.readthedocs.io/en/latest/user-management.html))
+* Deprecated `Kreait\Firebase\Factory::withServiceAccount()`, use `Kreait\Firebase\Factory::withServiceAccountAndApiKey()` instead 
+* Deprecated `Kreait\Firebase::asUserWithClaims()`, use `Kreait\Firebase\Auth::getUser()` and `Kreait\Firebase::asUser()` instead
+* Deprecated `Kreait\Firebase::getTokenHandler()`, use `Kreait\Firebase\Auth::createCustomToken()` and `Kreait\Firebase\Auth::verifyIdToken()` instead.
+* Added migration instructions for deprecated methods, see [Documentation](https://firebase-php.readthedocs.io/en/latest/migration.html#to-3-2)
+  
 ## 3.1.2 - 2017-08-11
 
 * Removed the restriction to the google/auth package to versions <1.0
--- a/docs/authentication.rst
+++ b/docs/authentication.rst
@@ -11,17 +11,25 @@ a user id as the first parameter, and an optional array with claims as the secon
 
 .. code-block:: php
 
-    use Kreait\Firebase;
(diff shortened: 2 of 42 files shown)


P110:
getUser works nicely but the contents are private and difficult to work with, is this the way it was intended or is that being changed? Just wondering, thanks for this update!

jeromegamez:
@P110 At the moment, yes, until I have figured out how to expose the user profile and the linked provider info in a comfortable way. Until then you should be able to extract any data from the IdToken directly.

P110:
For now, on a slightly related note, I'm getting a token from a login provider and then using verifyIdToken to get the user ID & object back to manipulate and carry on. The deprecation message tells me to use createCustomToken() instead but that does the opposite of what I'm trying to do?? Is anything new coming to convert a token into a user ID or is this feature deprecated entirely?

P110:
Quick update: For testing purposes, I've just made the idToken part of the user class public rather than private so I can work with it.

jeromegamez:
@P110 There already is a public `getIdToken()` method on the user class. Have you updated the library based on this branch since your last comment so that you can use the ID verification methods that were added since then?

P110:
Sorry, hadn't noticed that, yeah I updated all seems to work well, thanks for re-including the idToken verification

Issue #120, opened 2017-10-06
Cannot use Kreait\Firebase\Http\Auth as Auth because the name is already in use 


mirahost:
#### 1. Which version of `kreait/firebase-php` are you using?

3.2.0

#### 2. Which version of PHP are you using?

7

#### 3. What's the issue?

When getDatabase is called it throws the following error:
Compile Error - Cannot use Kreait\Firebase\Http\Auth as Auth because the name is already in use in DOCROOT/vendor/kreait/firebase-php/src/Firebase/Database.php on line 10

#### 4. Code that lead to the issue
```

        $serviceAccount = \Kreait\Firebase\ServiceAccount::fromArray($config);
        $firebase = (new \Kreait\Firebase\Factory)
            ->withServiceAccountAndApiKey($serviceAccount, $apiKey)
            ->withDatabaseUri('https://example.firebaseio.com')
            ->create();

        $database = $firebase->getDatabase();
```

Result:

Compile Error - Cannot use Kreait\Firebase\Http\Auth as Auth because the name is already in use in DOCROOT/vendor/kreait/firebase-php/src/Firebase/Database.php on line 10

jeromegamez:
I can't reproduce this error, please make sure that you have no `use Kreait\Firebase\Http\Auth` or `use Kreait\Firebase\Http\Auth as Auth` statement elsewhere in your code.

asahasrabuddhe:
@mirahost are you using this piece of code in conjunction with Laravel's Auth facade?

renddrew:
I'm getting the same error in an otherwise empty project. Installed with composer.

jeromegamez:
@mirahost @renddrew I still can't reproduce the issue, here is what I did:

```bash
mkdir fb-issue-120
cd fb-issue-120
composer require kreait/firebase-php
```

Then I created a `test.php` file:

```php
<?php

require __DIR__.'/vendor/autoload.php';

use Kreait\Firebase;

$serviceAccount = Firebase\ServiceAccount::fromJsonFile('/path/to/google-service-account.json');
$apiKey = '<Firebase Web API key>';

$firebase = (new Firebase\Factory)
    ->withServiceAccountAndApiKey($serviceAccount, $apiKey)
    ->create();

$database = $firebase->getDatabase();
```

The execution of `php test.php` then returned no error.

ifilmstuff:
[...]

PR #125, opened 2017-10-30
Rename interface to prevent namespace collisions

bennyt2:
This PR proposes a fix for issue #120.

For some reason, the 'Auth' interface name is colliding with another
'Auth' class or interface in code. This commit resolves that issue
by renaming the 'Auth' interface to 'AuthInterface'.

### DIFF
--- a/src/Firebase.php
+++ b/src/Firebase.php
@@ -12,7 +12,7 @@ use Kreait\Firebase\Auth\User;
 use Kreait\Firebase\Database;
 use Kreait\Firebase\Database\ApiClient;
 use Kreait\Firebase\Exception\LogicException;
-use Kreait\Firebase\Http\Auth;
+use Kreait\Firebase\Http\AuthInterface;
 use Kreait\Firebase\Http\Auth\CustomToken;
 use Kreait\Firebase\Http\Auth\UserAuth;
 use Kreait\Firebase\Http\Middleware;
@@ -150,7 +150,7 @@ class Firebase
         return $this->tokenHandler;
     }
 
-    private function withCustomAuth(Auth $override): Firebase
+    private function withCustomAuth(AuthInterface $override): Firebase
     {
         $firebase = new self($this->serviceAccount, $this->databaseUri, $this->tokenHandler);
         $firebase->database = $this->createDatabase()->withCustomAuth($override);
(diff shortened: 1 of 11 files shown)


jeromegamez:
Although I can't reproduce this problem, I do realize that having two classes with the same name but a different purpose is not really nice, so thank you for taking care of this, @bennyt2.

Could you please change the interface name to `AuthenticationMethod` (without the `Interface` suffix)? I think this would be a good opportunity to make the interface name more meaningful.

bennyt2:
@jeromegamez Thanks for your feedback, and I agree, `AuthenticationMethod` is much better than `AuthInterface` :).

I can't seem to get the Travis build to work. It says `github-oauth.github.com is not defined.` Let me know if there's something I need to do on my end to make this work.

jeromegamez:
The Travis CI build currently is broken for PRs in general, I will fix this later - as long as the test suite runs on your and my computer, we're safe - and it does on mine 🎉 .

There's just one thing I realized while looking over it again... changing the interface name is a BC break, so strictly speaking, I can not release this as a minor or patch release. If someone uses the `Firebase::withCustomAuth()` or `Database::withCustomAuth()` methods in their projects, this will break their project :(.

On the other hand - nobody *should* have needed to use those methods directly: Firebase only supports the authentication methods provided in the library, so there should have been no need to add an additional custom authentication method...

But strictly speaking… 

I have to think about this, perhaps there's another way to fix #120…

jeromegamez:
@bennyt2 I'm really bummed about the BC break thing (see https://github.com/kreait/firebase-php/issues/120#issuecomment-340705955) - on the positive side PR builds should now work again, so if you rebase your branch with the current master branch, the checks should be green again.

bennyt2:
@jeromegamez Thanks for fixing the Travis PR issue. The build is now passing.

Great catch on the BC problem. I will think about other ways to address the issue too. Let me know if you think of anything :)

jeromegamez:
@bennyt2 The issue should be fixed with [v3.3.3](https://github.com/kreait/firebase-php/releases/tag/3.3.3), so I will close this PR now. Thank you for the time and effort you've put into it and for helping me out in locating the issue and for drawing my attention to the broken Travis CI builds, I greatly appreciate it! 🌺

Issue #132, opened 2017-11-22
Can't write to .settings/rules.json


bennyt2:
#### 1. Which version of `kreait/firebase-php` are you using?

3.4.2

#### 2. Which version of PHP are you using?

PHP 7.0.8

#### 3. What's the issue?

I cannot PUT rules to .settings/rules.json. [This help document](https://firebase.google.com/docs/database/rest/app-management) states that the URL is valid, but I get the following error when trying to do so.

```
[Kreait\Firebase\Exception\InvalidArgumentException]                                 
The child key ".settings" contains one of the following invalid characters: ".$#[]"  

Exception trace:
 () at /srv/parser/vendor/kreait/firebase-php/src/Firebase/Database/Reference/Validator.php:64
 Kreait\Firebase\Database\Reference\Validator->validateChars() at /srv/parser/vendor/kreait/firebase-php/src/Firebase/Database/Reference/Validator.php:31
 Kreait\Firebase\Database\Reference\Validator->validateUri() at /srv/parser/vendor/kreait/firebase-php/src/Firebase/Database/Reference.php:47
 Kreait\Firebase\Database\Reference->__construct() at /srv/parser/vendor/kreait/firebase-php/src/Firebase/Database.php:70
 Kreait\Firebase\Database->getReference() at /srv/parser/src/Path/To/Bundle/API/PutRequest.php:31
 Path\To\Bundle\API\PutRequest->execute() at /srv/parser/src/Path/To/Bundle/Service/RuleBuilder.php:79

```

#### 4. Code that led to the issue

```
$response = $this->client->getDatabase()->getReference($path)->set($data);

$client = an implementation of Kreait\Firebase.
$path = '.settings/rules';
$data = '{"users":{"$uid":{".read":"auth != null && auth.uid == $uid",".write":"$uid === auth.uid"}},".read":true,".write":false,"testA":{"list":{".indexOn":["cd"]},"hello":{".indexOn":["he","hr","hh"]}}}';
```

#### 5. Thoughts

In Kreait\Firebase\Database\Reference\Validator->validateUri(), could we make an exception for the ".settings/rules" path? I can't find other paths in the API that have a "." in front, but I could be missing something.

jeromegamez:
Closed with #136 :)

PR #133, opened 2017-11-22
Allow valid .settings/rules PUT request

bennyt2:
My attempt at solving issue #132.

Per [the docs](https://firebase.google.com/docs/database/web/structure-data), keys "cannot contain ., $, #, [, ], /, or ASCII control characters 0-31 or 127". However, [writing rules through the REST API](https://firebase.google.com/docs/database/rest/app-management) requires using the '.settings/rules' key, which contains an invalid character. This commit makes a special exception for the '.settings/rules' key.

### DIFF
--- a/src/Firebase/Database/Reference/Validator.php
+++ b/src/Firebase/Database/Reference/Validator.php
@@ -20,15 +20,25 @@ class Validator
      */
     public function validateUri(UriInterface $uri)
     {
+        // Firebase has valid REST URL endpoints that contain invalid
+        // characters. For example, the PUT .settings/rules endpoint
+        // writes rules to the database. This array allows us to filter
+        // validation on those technically valid URLs.
+        $validUrls = [
+            '.settings/rules',
+        ];
+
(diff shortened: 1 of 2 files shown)


jeromegamez:
Thank you for your contribution, Ben! Unfortunately, while being part of the Firebase REST API, this endpoint doesn't represent a Reference - you can only PUT (and probably GET) the database rules, but not push children or apply filters on it. So I think treating the rules as "just another reference" has the risk of introducing side effects, even if it's only future Github issues like "Why can't I update a Firebase database rule in a child key, my IDE autocompletion told me I could" :)

I'm not at my computer right now, so I can't check the library code in detail, but what would you think about something like `$firebase->getDatabase()->getRules()/->setRules()` to retrieve/update the rules?

bennyt2:
@jeromegamez Thanks for your quick reply, and you make a valid point! I like your suggestion. I'll attempt to do that and create a new PR.

bennyt2:
See PR #136

PR #135, opened 2017-11-24
prevent throw array access error when `orderByChild('nested/key')`

llstarscreamll:
This PR prevent throw array access errors when use nested Firebase paths on `orderByChild`. Look at the next query example:

```php
$firebase->getReference('users')
  ->orderByChild('profile/email') // nested Firebase path
  ->startAt('john@')
  ->getValue();
```

The schema:

```json
"users": {
  "a1": {
    "profile": {
      "email": "john@doe.com"
     }
  }
}
```
Before this throwed `[ErrorException] Undefined index: profile/email`.

Now this query doesn't throw exceptions.

### DIFF
--- a/src/Firebase/Database/Query/Sorter/OrderByChild.php
+++ b/src/Firebase/Database/Query/Sorter/OrderByChild.php
@@ -31,7 +31,7 @@ final class OrderByChild implements Sorter
         $childKey = $this->childKey;
         
         uasort($value, function ($a, $b) use ($childKey) {
-            return ($this->get($a, $childKey) ?? null) <=> $this->get($b, $childKey) ?? null;
+            return ($this->accessToArrayPath($a, $childKey) ?? null) <=> $this->accessToArrayPath($b, $childKey) ?? null;
         });
         
         return $value;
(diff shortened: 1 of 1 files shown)


jeromegamez:
Thank you for the contribution! I didn't even know it was possible to order by a nested child value - is it? And if so, why isn't it documented in the official docs 😅?

I will have a deeper look at it shortly - I will certainly rework the helper method, because I think it could be structured a lot nicer (and code climate agrees 😅), but I will make sure that your commits are preserved.

llstarscreamll:
Hi @jeromegamez, I have been very busy these days, I made a small cleaning up and renamed the helper method, all is green now, make the modifications you consider to this PR.

Thank you so much for this library!!

Cheers!!

jeromegamez:
Thank you Johan! The reason why it's taking me so long to review and integrate this feature is that I have to understand what's happening and to be sure that it's the best and easiest possible way to provide a feature (as it's me who has to provide support for the library in the future :)).

I do understand what's happening and why this is needed, but my gut tells me that there's an easier way to do it ^^. Your PR will be merged and you'll get attributed for the feature, please allow me a little bit more time for it :). Cheers!

jeromegamez:
Hey Johan! Thanks again for your contribution - please have a look at [Release 3.7.0](https://github.com/kreait/firebase-php/releases/tag/3.7.0) - it includes your contribution and the changes I made based upon your work:

- Added tests to ensure that everything is working as expected (it did with your solution)
- Used [Jmespath](https://github.com/jmespath/jmespath.php) to check for the nested values - the library is already included in the SDK and provides a similar functionality to the one you provided.

Cheers!

llstarscreamll:
Great!! Thank you so much!!

PR #136, opened 2017-11-24
Get and Set Firebase Rules

bennyt2:
Get and Set Firebase Rules via the REST API, as noted in this documentation: https://firebase.google.com/docs/database/rest/app-management

This is a proposed fix for Issue #132.

### DIFF
--- a/docs/realtime-database.rst
+++ b/docs/realtime-database.rst
@@ -378,6 +378,34 @@ includes the sent request and the received response object:
     }
 
 
+*****
+Rules
+*****
+
+Manage Firebase Realtime Database Rules through the following examples. 
+
+Detailed information can be found on
+`the official Firebase documentation page for Rules management via REST <https://firebase.google.com/docs/database/rest/app-management>`_
+
+Update Rules
+=============
+
+.. code-block:: php
+
+        $this->firebase->getDatabase()->setRules([
+            'rules' => [
+                '.read' => true,
+                '.write' => false
+            ]
+        ]);
+
+Retrieve Rules
+=============
+
+.. code-block:: php
+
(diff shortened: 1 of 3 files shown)


jeromegamez:
Ooooh, this looks nice, I like it a lot already. As soon as I get back to my computer, I will check it out and play it through once.

Would you mind adding a small section to the docs as well? (The incoming merge doesn't depend on it :))

bennyt2:
Docs added. I think I did them right. I had trouble installing Sphinx on my machine, so I wasn't able to review the formatting locally. Let me know if they need to be changed or improved.

jeromegamez:
Thank you Ben! There are some minor issues, but nothing serious. I'll rewrite the docs a little, and I will also add some try/catch blocks to the code - for example: if the first `rules` key is missing, the API returns an error, which is currently not catched. It would perhaps make sense to abstract the first field away. So the behaviour will probably change slightly.

Anyways, I will merge your changes and add mine afterwards. Thank you very much!

bennyt2:
@jeromegamez Thank you for merging this code, and for your speedy and thoughtful replies. Have a great weekend!

jeromegamez:
@bennyt2 I have made some (yet unreleased) changes to the code and the documentation, you can see them here:

Commit: https://github.com/kreait/firebase-php/commit/c2128d61c3741760a6dcb13b9b089adb7f61e565
Docs: http://firebase-php.readthedocs.io/en/latest/realtime-database.html#database-rules

bennyt2:
@jeromegamez I tested this out in my app and it works great. I like that you made the RuleSet class. Makes it much easier to visualize, and forces people to use Rules correctly. Thank you!!

Issue #95, opened 2017-06-20
error in factory.php


wardberckmans:
In Factory.php (/firebase-php/src/Firebase/Factory.php) of package version 3.1.1, I get an error on line 37. The error says: parse error, expecting `';'' or `'{''. This is on the line with "public function withCredentials(string $credentials): self"
I did not manually change this file so I have no clue where this error is coming from.

jeromegamez:
Line 37 in the Factory is a comment (see https://github.com/kreait/firebase-php/blob/3.1.1/src/Firebase/Factory.php#L37).

My best guesses are that you are either not on PHP 7 or something went wrong when implementing the library.

The tests on Travis CI would have failed if there actually was a syntax error (https://travis-ci.org/kreait/firebase-php/builds/244075870).

Issue #75, opened 2017-04-13
Syntax errors found in current version


cssjockey:
syntax error, unexpected ':', expecting ';'  ... 
vendor/kreait/firebase-php/src/Firebase.php on line 24

iamtankist:
duplicate of #66

cssjockey:
Thanks

Issue #66, opened 2017-02-28
Errors in code


drpain:
1350#0: *2098534 FastCGI sent in stderr: "PHP message: PHP Parse error:  syntax error, unexpected ':', expecting ';' or '{' in /var/www/vendor/kreait/firebase-php/src/Firebase.php on line 23"

Really should have error_reporting enabled man

jeromegamez:
@drpain Your error is a result of using the library with PHP < 7.0, see [Return type declarations in the PHP manual](http://php.net/manual/en/functions.returning-values.php#functions.returning-values.type-declaration).

If you're using PHP 5.x, the [1.x release](https://github.com/kreait/firebase-php/tree/1.x) might fit your needs.

Really should have read the readme man 😛

drpain:
Weird because my severs are running PHP 7. I will have to look again, but caused 30minutes of downtime so I would have to do it carefully.

jeromegamez:
@SYK360 Be assured, if you get a syntax error, the environment you're running the script in is *not* running PHP 7.x.

You can check this by adding the line `echo phpversion(); exit` somewhere in your script and run it on the environment you're getting the error with.

naeimrezaeian:
how can run code on php under 7.x ?

drpain:
I have personaly bumped my php version to 7.2 since this will be the new LTS for php from the end of 2018, and I am now using the library with great success. 

Example of my dockerfile for reference.
https://github.com/drpain/DockIt/blob/master/images/php-fpm/Dockerfile

jeromegamez:
@naeimrezaeian If you have to use a PHP version < 7.0, you can try the 1.2.2 version of this SDK: https://firebase-php.readthedocs.io/en/1.2.2/ - but it has limited features, and I can't provide support for it.

PR #144, opened 2018-01-05
[docs] overview - do not require specific version, let composer pick one

TomasVotruba:


jeromegamez:
I added the version on purpose to reduce frequent issues reported by people requiring the SDK in an unsupported PHP environment, e.g. https://github.com/kreait/firebase-php/issues/66, https://github.com/kreait/firebase-php/issues/75, https://github.com/kreait/firebase-php/issues/95.

When using PHP 5.x, a `composer require kreait/firebase-php` would install version 1.x, so I prefer keeping the constraint in the documentation.

Thank you for the PR nonetheless, Tomas, I do really appreciate that you took the time to create it!

:octocat:

TomasVotruba:
@jeromegamez How can it be installed in wrong PHP version? That should be handled in `composer.json`, right?

jeromegamez:
Composer will check the current PHP version and install the latest version supported for that PHP version - as Release 1.x supports PHP ^5.5 (see https://packagist.org/packages/kreait/firebase-php#1.2.2), composer would use that one.

TomasVotruba:
And Relase 1.1 doesn't support PHP 5.5? So `composer.json` is invalid there?

jeromegamez:
Release 1.1 does support PHP 5.5. The point is: depending on the PHP version `composer require kreait/firebase-php` will not always install the same version. `composer require kreait/firebase-php ^3.0` however will install always install version 3.x (on PHP ^7.0) or fail (on PHP <7) with the message that this PHP version is not supported.

TomasVotruba:
I see, the documentations is out of sync.

Now it makes sense, thanks :)

PR #151, opened 2018-01-16
Ensure correct ID token verification

jeromegamez:
:octocat:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## Unreleased
+
+* Bugfix: Ensure that ID tokens are verified fully and completely (discovered by [@hernandev](https://github.com/hernandev), thanks!)
+
 ## 3.8.0 - 2018-01-12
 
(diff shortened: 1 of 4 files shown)


jeromegamez:
@hernandev Could you please have a look at this?

hernandev:
are you using the child package or imported the code into this one?

hernandev:
Could not find where the Token\Domain is comming from

jeromegamez:
Yes, I'm using the [kreait/firebase-tokens-php](https://github.com/kreait/firebase-tokens-php): https://github.com/kreait/firebase-tokens-php/blob/master/src/Domain/Verifier.php

It's the Verifier interface, I then inject the implementation in the factory.

hernandev:
@jeromegamez I'm good then. gonna drop you an email now, if that's ok.

Thanks for such a fast fix.

hernandev:
@jeromegamez I'm available to finish it up testing after upgrade to ensure it's all good.

Your email is not public, so just say hi at diego@hernandev.com just for me to tell you some quick tips

PR #153, opened 2018-01-16
update method to accept either token as string or Token class instance

### LINKED ISSUE #152: verify method now expects token instead of  string
#### 1. Which version of `kreait/firebase-php` are you using?

`3.8.1`

#### 3. What's the issue?
Introduced on 3.8.1, `$auth->verifyTokenId()` method id no longer expecting a string but a `\Lcobucci\JWT\Token` instance.

----
I'll provide a PR for it on the next couple hours.

Thanks

hernandev:
This PR allows for `verifyIdToken` method on `Auth` class to accept both a string or a `Token` instance.

Fixes #152

### DIFF
--- a/src/Firebase/Auth.php
+++ b/src/Firebase/Auth.php
@@ -7,6 +7,7 @@ use Kreait\Firebase\Auth\CustomTokenGenerator;
 use Kreait\Firebase\Auth\IdTokenVerifier;
 use Kreait\Firebase\Auth\User;
 use Kreait\Firebase\Util\JSON;
+use Lcobucci\JWT\Parser;
 use Lcobucci\JWT\Token;
 use Psr\Http\Message\ResponseInterface;
 
@@ -138,6 +139,10 @@ class Auth
 
     public function verifyIdToken($idToken): Token
     {
+        if (is_string($idToken)) {
+            $idToken = (new Parser())->parse($idToken);
+        }
+
         return $this->idTokenVerifier->verify($idToken);
     }
 


jeromegamez:
Are you sure? The used Verifier (https://github.com/kreait/firebase-tokens-php/blob/master/src/Verifier.php#L41) parses an ID token string already, and I couldn't recreate the issue.

Just to be sure, I added unit and integration tests with b3bd43553af7a092798868d16179960f266b5550 and f370001613bd5702817227f40916c502184a6d13, could you please rebase your branch and check if this really doesn't work for you?

You can execute the unit test suite with `make tests` or `vendor/bin/phpunit --testsuite unit`.

hernandev:
Could be some caching, I'll get back to you.

hernandev:
This is the error:
```
TypeError: Argument 1 passed to Kreait\Firebase\Exception\Auth\InvalidIdToken::__construct() must be an instance of Lcobucci\JWT\Token, string given, called in /var/www/app/vendor/kreait/firebase-php/src/Firebase/Auth/IdTokenVerifier.php on line 33
```

hernandev:
Ok, you got the line, after a exception is raised, the string is being passed to the Exception instance, which does not know how to handle strings.

That's why this PR fixed it, but by the wrong way since it will be trying to parse it on two places.

I would recommend you to parse as soon as possible on the flow to avoid a "is it already parsed" question everywhere.

So, gonna close this, you can solve with the best approach!

See ya!

PR #174, opened 2018-03-01
Add signing-in and updating-profile apis

ttskch:
This PR adds following two apis.

* [Sign in with email / password](https://firebase.google.com/docs/reference/rest/auth/#section-sign-in-email-password)
* [Update profile](https://firebase.google.com/docs/reference/rest/auth/#section-update-profile)

### DIFF
--- a/src/Firebase/Auth/ApiClient.php
+++ b/src/Firebase/Auth/ApiClient.php
@@ -161,6 +161,49 @@ class ApiClient
         ]);
     }
 
+    /**
+     * @param string $email
+     * @param string $password
+     *
+     * @see https://firebase.google.com/docs/reference/rest/auth/#section-sign-in-email-password
+     *
+     * @return ResponseInterface
+     */
+    public function signinWithEmailAndPassword(string $email, string $password): ResponseInterface
+    {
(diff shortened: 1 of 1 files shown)


jeromegamez:
Hey @ttskch, thank you for you PR!

I'm currently working on creating and updating users with all possible properties and hope to have this ready in a short time.

Concerning the "user sign in" functionality: I had this in the library in previous releases but deliberately removed them again, because the official Admin SDKs don't provide this either and because I felt it not being right in the scope of this library as this is an Admin SDK, and signing in with a user's credentials on a server backend doesn't feel "right" to me. In my opinion (at the moment), the user flow should be performed with the officially recommended ways, being: a client application handles the authentication, creates an IdToken and sends this to a backend.

I might consider adding this again in the future (either in this or in another library), but at the moment, I'd prefer not to.

Thanks again for the PR, although I will not merge it, I do appreciate the time you've put into it!

jeromegamez:
I have published the first steps of my implementation at https://github.com/kreait/firebase-php/pull/177 and I would be glad if you wanted to test this.

PR #187, opened 2018-04-03
Update wrong doc in authentication.rst

innaj:
asUser() must be called after create()  (only create() returns the Firebase Object)

### DIFF
--- a/docs/authentication.rst
+++ b/docs/authentication.rst
@@ -89,8 +89,8 @@ with the identifier you used to represent your service in your Security Rules.
 
     $firebase = (new Factory)
         ->withServiceAccount($serviceAccount)
-        ->asUser('my-service-worker')
-        ->create();
+        ->create()
+        ->asUser('my-service-worker');
 
 
 ***************************


jeromegamez:
Hey @innaj, are you sure you are on the latest (4.x) release of the library? The factory has a `asUser()` method, the `Firebase` object hasn't.

Factory: https://github.com/kreait/firebase-php/blob/master/src/Firebase/Factory.php#L86-L93
Firebase: https://github.com/kreait/firebase-php/blob/master/src/Firebase.php

innaj:
Uh, I was on the 3.9.x release, didn't know there is a 4.x branch...

jeromegamez:
The differences between 3.x and 4.x are not too big (I think/hope), and you can find the documentation for authenticating with limited privileges at http://firebase-php.readthedocs.io/en/latest/authentication.html#authenticate-with-limited-privileges - and thanks for the PR anyways - I saw this was your first one and I appreciate it!

Issue #117, opened 2017-10-04
Support Firestore


jeromegamez:
See https://firebase.googleblog.com/2017/10/cloud-firestore-for-rtdb-developers.html

aloisejulian:
Yess! we want support for firestore... Do we have a estimate date? Thanks!!

ChrisSmith5:
Any update on Firestore? Very excited for this new feature!

jeromegamez:
Unfortunately I cannot provide an ETA for new features - I am working on this library in my limited free time and (unfortunately 😅) quite picky/thorough with new code, so this will certainly take some time.

aloisejulian:
@jeromegamez thanks! i will wait for the release. Meanwhile i will use firebase

anelad:
Even you cannot give an exact ETA, can you say something like in 3 months or in 5 months? We will start to implement my project in 2 weeks, so I need to know if I need to do some changes on project's design. Thank you.

etopian:
here is a class to get you on your way.
https://gist.github.com/jasonbyrne/6b63e7d59c5ed97ef95ff2c78791cfae

P110:
+1 for this feature if possible!

P110:
Sidenote: For anyone wanting to do this before we get support in this SDK you might be able to use this, I'm going to try and use it later on today to see how stable it is.

riyuk:
The (huge) downside of the googlecloudplatform framework is that it requires the gRPC extension which you normally dont get on a shared hosting

jeromegamez:
Yep, that's why haven't looked deeper into that yet. Just exposing the googlecloudplatform implementation with `getFirestore()` in the `Firebase` class would be easy enough - but as you said, this would require everybody to install the gRPC extension.

On the other hand, I'm not really keen on re-implementing everything here 😅.

On the other other hand it should also be not too much of a problem to use both in tandem.

morrislaptop:
As discussed in #198, Firestore is now supported in a separate package at https://github.com/morrislaptop/firestore-php

Which is standing on the shoulder of giants of this package :)

@jeromegamez Could you add to the readme?

PR #198, opened 2018-05-20
Support Firestore

morrislaptop:
Supports creating and getting documents. 

Nowhere finished yet, but thought I would give a heads up that I'll be working on this as I need it for my current project. 

Addresses #117 

- [x] Get
- [x] Set
- [ ] Delete
- [ ] Add
- [ ] Transactions (beginTransaction, commit, rollback)
- [ ] Reference value support
- [ ] Batch Get
- [ ] List Documents
- [ ] Query
- [ ] Order
- [ ] Limit
- [ ] Indexes (create, delete, list, get)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,6 +21,7 @@
         "ext-openssl": "*",
         "giggsey/libphonenumber-for-php": "^8.9",
         "google/auth": "^0.11.0|^1.0",
+        "google/cloud-core": "^1.20",
         "guzzlehttp/guzzle": "^6.2.1",
         "kreait/firebase-tokens": "^1.1.1",
         "lcobucci/jwt": "^3.2",
@@ -29,8 +30,10 @@
     },
     "require-dev": {
         "friendsofphp/php-cs-fixer": "^2.0",
+        "morrislaptop/var-dumper-with-context": "^0.2.0",
         "phpstan/phpstan-phpunit": "^0.9.2",
-        "phpunit/phpunit": "^6.0|^7.0"
+        "phpunit/phpunit": "^6.0|^7.0",
+        "symfony/var-dumper": "^3.3"
     },
     "autoload": {
         "psr-4": {
--- a/src/Firebase.php
+++ b/src/Firebase.php
@@ -4,6 +4,7 @@ namespace Kreait;
 
 use Kreait\Firebase\Auth;
 use Kreait\Firebase\Database;
+use Kreait\Firebase\Firestore;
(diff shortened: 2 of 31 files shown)


morrislaptop:
@jeromegamez @aloisejulian @ChrisSmith5 @anelad @etopian @P110 @riyuk any thoughts?

jeromegamez:
Thanks for all the work you've already put into this, @morrislaptop, I really appreciate it!

However, as I said in https://github.com/kreait/firebase-php/issues/117#issuecomment-384626919, I am not keen of re-implementing of what's already being actively maintained in https://github.com/GoogleCloudPlatform/google-cloud-php-firestore, just to not have the requirement of the grpc extension.

Please keep in mind that once this PR would be merged, I would have to maintain it as well as keep track of changes in the copy-pasted classes (e.g. the ValueMapper class).

I currently see two solutions here:

- You could fork the PHP Admin SDK and maintain the Firestore part of it. I would then try to ensure that e.g. the factory and the Firebase class remain extensible, as well as link to your fork.

- We extend the factory and the Firebase class to allow instantiating/using a FirestoreStore instance with some `class_exists('\Google\Cloud\Firestore\FirestoreClient')` checks, and add `google/cloud-firestore` to the "suggest" section in the composer.json file and in the documentation.

I do want to keep the entrypoint to Firebase with PHP low by not requiring the grpc extension by default, but I don't think duplicating the complex Firestore functionality is the best solution for this, especially since the work to provide the functionality has already been done by Google.

What do you think?

jeromegamez:
The second solution (providing a gateway to `google/cloud-firestore` if the class exists) is something that's on my todo list already, and I would put it on the top of the list if you want to keep using this SDK, are ready to install the grpc extension and just need a way to get access to Firestore.

morrislaptop:
Hi @jeromegamez ,

I think the top goal is to avoid the grpc requirement. It is a massive barrier to entry.

So that just leaves the first solution. Just to be clear my approach will be:

* Creating a separate package `morrislaptop/firestore-php`
* Extend `Kreait\Firebase\Factory` to create a `Firestore` instance which is passed to:
* An extended `Kreait\Firebase` with a `getFirestore()` method

Users could then use simply use `Morrislaptop\Firestore\Factory` to interact with Firestore and all the other stuff this library provides. 

Sound like a plan?

jeromegamez:
Sounds like a plan!

morrislaptop:
WIP at https://github.com/morrislaptop/firestore-php

PR #200, opened 2018-05-23
Implement FCM topic management

jeromegamez:
Adds methods to subscribe and unsubscribe devices to and from topics.

### How to test

```php
<?php

declare(strict_types=1);

use Kreait\Firebase;
use Kreait\Firebase\Util\JSON;

require_once __DIR__.'/vendor/autoload.php';

$serviceAccount = Firebase\ServiceAccount::fromJsonFile('/path/to/credentials.json');

$firebase = (new Firebase\Factory())
    ->withServiceAccount($serviceAccount)
    ->create();

$messaging = $firebase->getMessaging();
$tokens = [
    // ...
];

echo JSON::prettyPrint($messaging->subscribeToTopic('foo', $tokens));
echo JSON::prettyPrint($messaging->unsubscribeFromTopic('foo', $tokens));
```

### Todo

- [x] Add tests
- [x] Add documentation

PR #201, opened 2018-05-23
Make functions protected instead of private so class can be extended

morrislaptop:
For the plan conceived at #198

### DIFF
--- a/src/Firebase/Factory.php
+++ b/src/Firebase/Factory.php
@@ -41,7 +41,7 @@ class Factory
     /**
      * @var string|null
      */
-    private $uid;
+    protected $uid;
 
     /**
      * @var array
@@ -109,7 +109,7 @@ class Factory
         return $this->serviceAccountDiscoverer ?? new Discoverer();
     }
 
-    private function getServiceAccount(): ServiceAccount
+    protected function getServiceAccount(): ServiceAccount
     {
         if (!$this->serviceAccount) {
             $this->serviceAccount = $this->getServiceAccountDiscoverer()->discover();
@@ -138,7 +138,7 @@ class Factory
         return sprintf(self::$storageBucketNamePattern, $serviceAccount->getProjectId());
     }
 
-    private function createAuth(): Auth
+    protected function createAuth(): Auth
     {
         $serviceAccount = $this->getServiceAccount();
 
@@ -153,7 +153,7 @@ class Factory
         );
     }
 
-    private function createDatabase(): Database
+    protected function createDatabase(): Database
     {
         $http = $this->createApiClient($this->getServiceAccount());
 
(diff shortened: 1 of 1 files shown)


jeromegamez:
I went one step further and changed all properties and methods from `private` to `protected` in https://github.com/kreait/firebase-php/commit/69f6f3e15acae01fd16c91697b0cc76de66259af

On a side note: please compose commit messages like described in https://chris.beams.io/posts/git-commit/ (here: present tense) so I don't have to bug you to rewrite them or change them on my own when I then manually merge ^^

morrislaptop:
Ok thanks! Need 2 more from https://github.com/morrislaptop/firebase-php/commit/9f79b24b199efd70de6eea87c1b0b9decbc2e813

Although it would be better if you added the below into ApiException.php

```php
        if (is_array($message)) {
            $message = $message['message'];
        }
```

jeromegamez:
Everything besides the factory should be implementable in your extension SDK without having to change the code of the main SDK.

For example, I would advise against reusing/extending the Database ApiClient for access to Firestore, as the calls to the Firestore API are quite different.

You also should not re-use/expose the `Kreait\Firebase` exceptions - your library should have/use its own namespaced exceptions - if you throw `Kreait\Firebase` exceptions instead of `Morrislaptop\Firestore` exceptions, people will come to my repo instead of yours :D

morrislaptop:
Ok will do. 

Can you publish a new release with the extensible stuff?

jeromegamez:
Will do! On a sidenote: I would recommend to not extend the Firebase class here https://github.com/morrislaptop/firestore-php/blob/master/src/Firebase.php , but to reimplement the getters as they're simple enough. 

The reason for the suggestion: I'm thinking about rewriting the class, also to be able to add an optional integration with Google's (grpc requiring) Firestore implementation, which will then probably be type hintet. Implementing the class separately while keeping the method names would prevent your lib from breaking once I do this.

An alternative approach could be to add a method like `createFirestore()` to your factory and only use that. Thinking about it... what do you think about removing the dependency from the "main" SDK totally and copy paste only the factory methods you need to create a Firestore instance? You would then be independent/safe from changes  in the Firebase SDK.

Either way, I will update the documentation and create a new release in the next 1, 2 days.

morrislaptop:
I've removed the `Firebase` class as that wasn't really doing much. 

I started to decouple the `Factory` but I don't think this is a good idea. There's a lot of good stuff with the service account discovery etc.. I would like to keep.

jeromegamez:
@morrislaptop The changes in the documentation are coming with 215d49f9c34080f0daa2e8126670b7bd1088605a - I will create a new release once I finished with #200

jeromegamez:
Merged and released with [4.8.0](https://github.com/kreait/firebase-php/releases/tag/4.8.0) 🎉

PR #207, opened 2018-06-11
Improve inheritance for type-hinting messages

artaex:
I ran into a problem passing around `Message` instances in my code. All three message types (`MessageToTopic`, `MessageToRegistrationToken`, and `ConditionalMessage`) implement the `Message` interface, but that interface does not guarantee access to properties like the notification. Therefore I suggest replacing the interface by an abstract class that implements the `MessageTrait`.

### DIFF
--- a/src/Firebase/Messaging/ConditionalMessage.php
+++ b/src/Firebase/Messaging/ConditionalMessage.php
@@ -6,10 +6,8 @@ namespace Kreait\Firebase\Messaging;
 
 use Kreait\Firebase\Exception\InvalidArgumentException;
 
-class ConditionalMessage implements Message
+class ConditionalMessage extends Message
 {
-    use MessageTrait;
-
     /**
      * @var Condition
      */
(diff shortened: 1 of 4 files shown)


jeromegamez:
Hey @artaex, thank you for taking the time to think about this and provide this PR.

You are right that the message interface is just a marker interface and doesn't help with type hinting in IDEs, this is due to the fact that I wanted to ship the feature as fast as possible in the beginning and finalize it in following releases (which is still my plan).

Unfortunately, changing the interface to an abstract class (or adding methods to the interface) is a breaking change - if someone has written a custom message type (`class MyCustomMessage implements Message`), their code would break. 

This means that the issue you're describing is unfortunately not solvable until the next major release of the SDK (which will be a rewrite anyways). I hope you will still continue using it :).

On a sidenote: I am very much in favor of composition over inheritance, so I try to avoid abstract classes as much as possible 😅.

Issue #212, opened 2018-06-19
hasChild fails for keys starting with string integers or hyphens


AndreasMaros:
Hi there! First things first, magnificent work! Incredibly useful project obviously executed with great enthusiasm.

#### 1. Which version of `kreait/firebase-php` are you using?
Just updated to the lastest 4.10.0 from 4.7..0. Problem persists.

#### 2. Which version of PHP are you using?
7.2.6 in development, 7.2.x in production

#### 3. What's the issue?
Kreait\Firebase\Database\Snapshot methods `hasChild()` and `getChild()` throw a `SyntaxError` somewhere down the line when using a (string) key not starting with an alphabetical character (?). 

Therefore it can't be used to check for or get UIDs starting with an integer or the IDs generated by `Kreait\Firebase\Database\Reference::push` (which seem to generally start with a hyphen `-`).

```
PHP Warning:  Syntax error at character 0
8tPF1tW8lobLQEmLdxsX9S6TWv82
^
Unexpected "number" token (nud_number). Expected one of the following tokens: "identifier", "quoted_identifier", "current", "literal", "expref", "not", "lparen", "lbrace", "flatten", "filter", "star", "lbracket"
```

(I suspect `JmesPath\search`, but mainly because I don't know what it does... :stuck_out_tongue: )

#### 4. Code that lead to the issue

```php
$uid = '8tPF1tW8lobLQEmLdxsX9S6TWv82';
$snap = $this->database->getReference('/path')->getSnapshot();
$exists = $snap->hasChild($uid);
```

jeromegamez:
Hey @AndreasMaros, thank you for the kind words and for taking the time to report the issue - you're apparently the first one to actually use a Snapshot directly without just fetching the value 😅.

I created #213 to address the problem and hope to get to it soon.

AndreasMaros:
Yeah, I'm a bit weird like that. But luckily that also means it was very easy to work around this issue :slightly_smiling_face:

jeromegamez:
Here's the new release with the included fix: [4.10.1](https://github.com/kreait/firebase-php/releases/tag/4.10.1) 🎉 

Thanks again! 🌺

AndreasMaros:
No – thank *you*. :1st_place_medal:

PR #213, opened 2018-06-19
Support children with non-alphabetical keys

jeromegamez:
Addresses #212


jeromegamez:
@AndreasMaros Could you please check if the changes in this branch fix the issue for you?

AndreasMaros:
That does indeed do the trick! Thank you, that was quick.

PR #220, opened 2018-07-13
Email Verification and Password Reset support continueUrl

wadewomersley:
Adds support for continueUrl to reset password and verify email emails that go out as per https://firebase.google.com/docs/auth/custom-email-handler

### DIFF
--- a/src/Firebase/Auth/ApiClient.php
+++ b/src/Firebase/Auth/ApiClient.php
@@ -194,19 +194,21 @@ class ApiClient
      *
      * @return ResponseInterface
      */
-    public function sendEmailVerification(string $idToken): ResponseInterface
+    public function sendEmailVerification(string $idToken, string $continueUrl = null): ResponseInterface
     {
         return $this->request('getOobConfirmationCode', [
             'requestType' => 'VERIFY_EMAIL',
             'idToken' => $idToken,
+            'continueUrl' => $continueUrl,
         ]);
     }
 
-    public function sendPasswordResetEmail(string $email): ResponseInterface
+    public function sendPasswordResetEmail(string $email, string $continueUrl = null): ResponseInterface
     {
         return $this->request('getOobConfirmationCode', [
             'email' => $email,
             'requestType' => 'PASSWORD_RESET',
+            'continueUrl' => $continueUrl,
         ]);
     }
 
--- a/src/Firebase/Auth/ApiClient.php
+++ b/src/Firebase/Auth/ApiClient.php
@@ -191,6 +191,7 @@ class ApiClient
 
     /**
      * @param string $idToken
+     * @param string $continueUrl
      *
      * @return ResponseInterface
      */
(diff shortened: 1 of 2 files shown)


src/Firebase/Auth/ApiClient.php
```diff
@@ -191,22 +191,25 @@ public function verifyPassword(string $email, string $password): ResponseInterfa
...
      *
      * @return ResponseInterface
      */
-    public function sendEmailVerification(string $idToken): ResponseInterface
+    public function sendEmailVerification(string $idToken, string $continueUrl = null): ResponseInterface
     {
         return $this->request('getOobConfirmationCode', [
             'requestType' => 'VERIFY_EMAIL',
             'idToken' => $idToken,
+            'continueUrl' => $continueUrl,
```
jeromegamez: Please wrap the array in `array_filter(...)` to avoid sending an empty `continueUrl`field if none is set (the small things :D)

src/Firebase/Auth/ApiClient.php
```diff
@@ -191,22 +191,25 @@ public function verifyPassword(string $email, string $password): ResponseInterfa
...
         ]);
     }
 
-    public function sendPasswordResetEmail(string $email): ResponseInterface
+    public function sendPasswordResetEmail(string $email, string $continueUrl = null): ResponseInterface
     {
         return $this->request('getOobConfirmationCode', [
             'email' => $email,
             'requestType' => 'PASSWORD_RESET',
+            'continueUrl' => $continueUrl,
```
jeromegamez: Please wrap the array in `array_filter(...)` to avoid sending an empty `continueUrl`field if none is set (the small things :D)

src/Firebase/Auth/ApiClient.php
wadewomersley: If I do that, am I OK to move the array to a var and add in the URL if not null instead (or use array filter instead of the `if` check below for future changes? Otherwise I'll have to add the array_filter inside the function *and* have a custom function so it doesn't remove an ID token of '0' which will look a bit messy?

```php
$data = [
    'requestType' => 'VERIFY_EMAIL',
    'idToken' => $idToken,
];

if ($continueUrl !== null) {
    $data['continueUrl'] = $continueUrl;
}

return $this->request('getOobConfirmationCode', $data);
```

jeromegamez: Wrapping it with `array_filter()` really is sufficient - a value of `0` would be rejected by the Firebase Auth API anyways.

wadewomersley: '0' is valid for Firebase.
image

wadewomersley: And that explains my confusion - this function has a token not a UID so by this point it would never be a UID. For some reason I had it in my head this function had the user ID!

jeromegamez:
Thanks @wadewomersley! I will update the documentation and create a new release soon.

PR #230, opened 2018-07-28
WIP: Allow the usage of the SDK on Google Cloud Engine

### LINKED ISSUE #229: Default service account on Compute Engine
I would like to use this library on Google's Compute Engine with the vm's default service account, without using a credentials json file.

Replacing ServiceAccountCredentials at https://github.com/kreait/firebase-php/blob/master/src/Firebase/Factory.php#L306 with [GCECredentials](https://github.com/google/google-auth-library-php/blob/master/src/Credentials/GCECredentials.php) works, but the library still needs a ServiceAccount to get the project id and the private key for signing custom tokens.

The project id can also be retrieved from the metadata service that GCECredentials uses: https://cloud.google.com/compute/docs/storing-retrieving-metadata

Signing custom tokens should be done using the iam signblob api call: https://firebase.google.com/docs/auth/admin/create-custom-tokens#using_a_service_account_id

jeromegamez:
Fixes #229

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,6 +24,7 @@
         "google/auth": "^0.11.0|^1.0",
         "guzzlehttp/guzzle": "^6.2.1",
         "kreait/firebase-tokens": "^1.1.1",
+        "kreait/gcp-metadata": "^1.0.1",
         "lcobucci/jwt": "^3.2",
         "mtdowling/jmespath.php": "^2.3",
         "superbalist/flysystem-google-storage": "^7.0"
--- a/src/Firebase/Auth/CustomTokenViaGoogleIam.php
+++ b/src/Firebase/Auth/CustomTokenViaGoogleIam.php
@@ -1,0 +1,74 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Kreait\Firebase\Auth;
+
+use Firebase\Auth\Token\Domain\Generator;
+use GuzzleHttp\ClientInterface;
+use GuzzleHttp\Exception\RequestException;
+use Kreait\Firebase\Exception\AuthException;
+use Kreait\Firebase\Util\JSON;
+use Lcobucci\JWT\Builder;
+use Lcobucci\JWT\Parser;
+use Lcobucci\JWT\Token;
+
+class CustomTokenViaGoogleIam implements Generator
+{
+    /**
+     * @var string
+     */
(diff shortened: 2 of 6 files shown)


nkovacs:
Test code:

```
        $uid = 'some-uid';
        $claims = [
            'premiumAccount' => true,
        ];
        $firebase = (new Factory)->create();
        $customToken = $firebase->getAuth()->createCustomToken($uid, $claims);
        echo "custom token: $customToken \n";
        $storage = $firebase->getStorage();
        $bucket = $storage->getBucket();

        $bucket->upload('test', ['name' => 'test.txt']);

        $objects = $bucket->objects();
        foreach ($objects as $object) {
            echo $object->name() . PHP_EOL;
        }
        $auth = $firebase->getAuth();
        $users = $auth->listUsers($defaultMaxResults = 1000, $defaultBatchSize = 1000);
        var_dump($users);
        $database = $firebase->getDatabase();
        $reference = $database->getReference('test');
        $value = $reference->getValue();
        var_dump($value);
```

Everything works except for the database:

```
Exception 'Kreait\Firebase\Exception\PermissionDenied' with message 'Unauthorized request.'
```

But it's possible I didn't set up the permissions correctly (the service account has Editor permissions, and the vm has full access to all cloud APIs).

Push notifications also work.

jeromegamez:
@nkovacs Wait, the scopes array isn't even used on GCE, so please don't test 😅 - I'll investigate further

jeromegamez:
@nkovacs I haven't yet found the time to set up my own GCE app - it would be great if you could check for the database permissions in the meantime to ensure that that's not it 😅

I don't know if this applies to GCE as well, but I stumbled upon this Medium article which hints on scopes having to be set/extended on a cloud instance: https://medium.com/google-cloud/firebase-developing-a-web-service-with-admin-sdk-flask-and-google-cloud-6fb97eb38b80

jeromegamez:
Implemented with 218028709912cab15c2905ce832a8cdd975a06e9

nkovacs:
Hi

Sorry for the late reply. Setting the scopes manually as described in the blog post you linked fixes the issue. Which is funny, because there doesn't seem to be any way to do it from the web ui (the vm already had full access to all cloud apis, but that wasn't enough).

Thanks!

PR #235, opened 2018-08-09
[CloudMessage] Adding getter for Notification and Data

valimero:


jeromegamez:
Could you please explain what this is for?

`CloudMessage` is meant to be filled with data, then to be passed to the Auth component which will send it to the Firebase Servers (after JSON encoding) - but not to be used besides that.

The new getters you're proposing would not change this behavior, and all data that the CloudMessage object receives is alrady known to you (because you filled it with that data).

Of course it doesn't hurt adding getters, but I need a better reason than a PR without a description :).

After that, I'll talk code ^^

valimero:
This is to use the data in another service and check if the data passed in CloudMessage is correct.

valimero:
How can we modify the MessageData after the CloudMessage was created ?

jeromegamez:
All objects in the SDK are immutable - once created you cannot change the message, but you can create a new message based on the previous one:

```php
<?php

declare(strict_types=1);

require_once __DIR__.'/vendor/autoload.php';

    'topic' => 'my-topic',
    'data' => ['key' => 'value']
]);

echo json_encode($message, JSON_PRETTY_PRINT);

$modifiedMessage = $message->withData([
    'key' => 'new_value',
    'another_key' => 'another_value',
]);

echo json_encode($modifiedMessage, JSON_PRETTY_PRINT);
```

will output

```
{
    "topic": "my-topic",
    "data": {
        "key": "value"
    }
}{
    "topic": "my-topic",
    "data": {
        "key": "new_value",
        "another_key": "another_value"
    }
}
```

jeromegamez:
If you can instantiate a CloudMessage object without errors, the data is correct - if you can instantiate a CloudMessage with invalid data, then it's a bug that should be fixed :)

But a `getData()` method would just return what you have given to it, the same for `getNotification()` so I don't see the benefits here.

If you absolutely need it, you can always wrap a CloudMessage in a class of your own - as long as your custom class implements `Kreait\Firebase\Messaging\Message`, it can be sent with the SDK

valimero:
The problem is when create a (new) modified message you can't retrieve the data in the old message.

In my controller I have only access to the CloudMessage object, not the data and the not notification inside.

jeromegamez:
I'm sorry (I really am) that I'm giving you such a hard time here - the thing is that each new public method adds to the things I have to maintain in the long term. That's why I insist so much on the separation of concerns...

The CloudMessage is not meant to be used outside of the scope of the SDK, and I should perhaps make that more clear in the code and in the documentation in general, I just assumed that this was clear. The limitation of public properties and methods enables me to act more freely in the "private" code in case that I need to change something.

If you need to dynamically modify the data that is used to generate a CloudMessage, I would recommend to pass around the *data* for a cloud message, and not a cloud message *containing* the data. The CloudMessage should only be built just before sending it via `$messaging->send($message);`.

You don't even need the CloudMessage, all you need is an array with data (see https://firebase-php.readthedocs.io/en/latest/cloud-messaging.html#sending-a-fully-configured-raw-message ) - you can modify and check your data then at will, and the SDK takes care that the message is valid and sends it.

Issue #202, opened 2018-05-27
Support for Session Cookies


KgotsoK:
#### 1. Which version of `kreait/firebase-php` are you using?
4.8.0

#### 2. Which version of PHP are you using?
7.1

#### 3. What's the issue?

Can we have support for Session Cookies:
`https://firebase.google.com/docs/auth/admin/manage-cookies`

It really simplifies Firebase Auth on stateful sites.

#### 4. Code that lead to the issue
N/A

jeromegamez:
Please don‘t think that I‘m ignoring your feature request - I‘ve already looked at a possible implementation but am hesitating a little because working with cookies adds a fair bit of complexity and I‘m trying to do this right from the start. Stay tuned :)

KgotsoK:
Thanks @jeromegamez . Glad to know you're playing with the idea. looking forward to it.

jeromegamez:
@KgotsoK @TheMrg @Slapbox @marcus-at-localhorst I finally got around to start working on this in #244 - it would be great if you could try it out and tell me if and how that works for you \o/

PR #244, opened 2018-09-13
Add support for Session Cookies

jeromegamez:
Addresses #202 

### Todo

- [x] Add functionality to create a session cookie
- [ ] Add functionality to verify a session cookie
- [ ] Write tests
- [ ] Write documentation

:octocat:

### DIFF
--- a/src/Firebase/Auth.php
+++ b/src/Firebase/Auth.php
@@ -3,16 +3,20 @@
 namespace Kreait\Firebase;
 
 use Firebase\Auth\Token\Domain\Generator as TokenGenerator;
-use Firebase\Auth\Token\Domain\Verifier as IdTokenVerifier;
+use Firebase\Auth\Token\Domain\Verifier as LegacyIdTokenVerifier;
 use Firebase\Auth\Token\Exception\InvalidSignature;
 use Firebase\Auth\Token\Exception\InvalidToken;
 use Firebase\Auth\Token\Exception\IssuedInTheFuture;
 use Kreait\Firebase\Auth\ApiClient;
+use Kreait\Firebase\Auth\IdTokenVerifier as NewIdTokenVerifier;
+use Kreait\Firebase\Auth\SessionTokenVerifier;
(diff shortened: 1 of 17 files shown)


RitcheyDevs:
Please is there any doc on how to use this yet?

src/Firebase/Auth.php
```diff
@@ -321,29 +347,45 @@ public function createCustomToken($uid, array $claims = null): Token
...
-            if (!$allowFutureTokens) {
-                throw $e;
-            }
-
-            $verifiedToken = $e->getToken();
-        }
+        if ($this->idTokenVerifier instanceof NewIdTokenVerifier) {
+            try {
+                $this->idTokenVerifier->verify($idToken);
+            } catch (Exception\InvalidToken $e) {
```
wallcrosstype: this catch InvalidToken hidden away exception message from inside verifier

for example, expired idToken will be thrown by verifyExpiry with message 'The token is expired since xxx'
with this too broad catch, will result in new InvalidToken without message from line 371

src/Firebase/Exception/ApiException.php
```diff
@@ -58,7 +57,7 @@ public function getResponse()
 
     private static function getTargetClassFromStatusCode($code): string
     {
-        if (\in_array($code, [StatusCode::STATUS_UNAUTHORIZED, StatusCode::STATUS_FORBIDDEN], true)) {
+        if (\in_array($code, [401, 403], true)) {
```
silvav: What is the reason to remove the usage of the constants?

jeromegamez: It's been quite some time, but I probably wanted to remove the transient dependency on the `fig/http-message-util` package (which is available through the inclusion of `kreait/firebase-tokens`). Removing that package would break this class.

I'm currently planning on a rework of both packages, so the issue might or might not apply in the future any more.

jeromegamez:
Postponed to after 5.0

PR #245, opened 2018-09-14
Changed CloudMessage properties privacy level from private to protected

garbalinskiy:
Making all properties private prevents from extending and simple child object initialization in __construct.
It would be nice not to use "private" everywhere and give more flexibility here.
For example would be possible to define classes like:  FriendIsOnlineNotification, NewMessageNotification and etc which would extend from CloudMessage class.

### DIFF
--- a/src/Firebase/Messaging/CloudMessage.php
+++ b/src/Firebase/Messaging/CloudMessage.php
@@ -11,32 +11,32 @@ class CloudMessage implements Message
     /**
      * @var MessageTarget
      */
-    private $target;
+    protected $target;
 
     /**
      * @var MessageData|null
      */
-    private $data;
+    protected $data;
 
     /**
      * @var Notification|null
      */
-    private $notification;
+    protected $notification;
 
     /**
      * @var AndroidConfig|null
      */
-    private $androidConfig;
+    protected $androidConfig;
 
     /**
      * @var ApnsConfig|null
      */
-    private $apnsConfig;
+    protected $apnsConfig;
 
     /**
      * @var WebPushConfig|null
      */
-    private $webPushConfig;
+    protected $webPushConfig;
 
     private function __construct()
     {


jeromegamez:
Hi @garbalinskiy, and thank you for taking the time to create this PR!

Making the constructors and properties private allows me to react to changing requirements and to change internal processes without breaking backwards compatibility. The SDK is meant to be used via its methods and ensures that all messages are passed through correctly and it doesn't prevent you from adding custom requirements around it.

On the other hand it protects me from feeling bad when I work on the code because I don't have to think about "What if someone extended the class X and I now break it for them?".

The CloudMessage is an implementation of the `Kreait\Firebase\Messaging` interface - if you want to create custom message types, you can implement the interface and use the `CloudMessage` as a builder.

These articles explain the "Composition over inheritance" approach better than I could:

- https://matthiasnoback.nl/2018/08/when-to-add-an-interface-to-a-class/
- https://matthiasnoback.nl/2018/09/final-classes-by-default-why/

jeromegamez:
Without knowing the specific requirement, here's an example of what I mean:

```php
use Kreait\Firebase\Messaging\Message;
use Kreait\Firebase\Messaging\MessageTarget;

class FriendIsOnline
{
    /**
     * @var string
     */
    private $friendName;

    /**
     * @var DateTimeInterface
     */
    private $onlineSince;

    public function __construct(string $friendName, \DateTimeInterface $onlineSince)
    {
        $this->friendName = $friendName;
        $this->onlineSince = $onlineSince;
    }

    public function asCloudMessageTo($recipient): Message
    {
        $friend = $this->friendName;
        $onlineSince = $this->onlineSince->format('H:i:s');

        return CloudMessage::withTarget(MessageTarget::TOKEN, $recipient)
            ->withNotification([
                'title' => "{$friend} just arrived!",
                'body' => "{$friend} has been online since {$onlineSince}"
            ]);
    }
}

$notification = new FriendIsOnline('Jérôme', new DateTime());

$firebase->getMessaging()->send($notification->asCloudMessageTo($deviceToken));
```

garbalinskiy:
But CloudMessage is not final and it means that you can still extend it and access jsonSerialize method because of \JsonSerializable interface.

Where jsonSerialize method access all private properties? Isn't it a little confusing? :)

garbalinskiy:
> Without knowing the specific requirement, here's an example of what I mean:
> 
> ```
> use Kreait\Firebase\Messaging\CloudMessage;
> use Kreait\Firebase\Messaging\Message;
> use Kreait\Firebase\Messaging\MessageTarget;
> 
> class FriendIsOnline
> {
>     /**
>      * @var string
>      */
>     private $friendName;
> 
>     /**
>      * @var DateTimeInterface
>      */
>     private $onlineSince;
> 
>     public function __construct(string $friendName, \DateTimeInterface $onlineSince)
>     {
>         $this->friendName = $friendName;
>         $this->onlineSince = $onlineSince;
>     }
> 
>     public function asCloudMessageTo($recipient): Message
>     {
>         $friend = $this->friendName;
>         $onlineSince = $this->onlineSince->format('H:i:s');
> 
>         return CloudMessage::withTarget(MessageTarget::TOKEN, $recipient)
>             ->withNotification([
>                 'title' => "{$friend} just arrived!",
>                 'body' => "{$friend} has been online since {$onlineSince}"
>             ]);
>     }
> }
> 
> $notification = new FriendIsOnline('Jérôme', new DateTime());
> 
> $firebase->getMessaging()->send($notification->asCloudMessageTo($deviceToken));
> ```

it looks like regular builder for Message and in this case we can even improve it with interface:
```
interface FriendIsOnlineMessageBuilder {
     public function build(): Message
}
```

jeromegamez:
Thinking about it... why would you need a FriendIsOnlineBuilder interface instead of just a MessageBuilder interface? It has just a build method (Simulation to Thema Message only having a jsonSerialize method). So you have different Implementations of "FriendIsOnline" Notifications?

garbalinskiy:
Honestly, you are right, MessageBuilderInterface should be enough :)

PR #271, opened 2018-12-19
Added example for webpush link

Shinhan:
I think an explicit example for a webpush link can be useful to others as it took me a while to find it in the docs.

### DIFF
--- a/docs/cloud-messaging.rst
+++ b/docs/cloud-messaging.rst
@@ -291,6 +291,9 @@ You can find the full WebPush configuration reference in the official documentat
             'body' => '$GOOG gained 11.80 points to close at 835.67, up 1.43% on the day.',
             'icon' => 'https://my-server/icon.png',
         ],
+        'fcm_options' => [
+            'link' => 'https://my-server/some-page',
+        ],
     ]);
 
     $message = $message->withWebPushConfig($config);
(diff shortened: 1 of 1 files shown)


jeromegamez:
Thanks for taking the time to create this PR! 

What is a WebPush Link? Where in the docs did you find it? If it's something that I can support in code, I will add it including your example - otherwise I would rather keep the links to the official documentation where devs can find every option, and not just select ones.

Shinhan:
WebPush link is used so if you click anywhere on the notification (if you are not at the website in question) the browser will open and navigate to the URL you have provided here. It has to be absolute URL of course, and is a feature we often use.

As for where I found it, the link provided in this very file has it mentioned but you have to scroll down a bit. Look for "The link to open when the user clicks on the notification. For all URL values, HTTPS is required." 

I did test this and it works fine, but I had to change our frontend code since the syntax is different from the legacy version (it used to be notification.click_action and is now fcmOptions.link)

Other documentations either don't talk about webpush at all or skip over the link functionality.

jeromegamez:
Thanks for the clarification and explanation, really appreciated and useful!

Merged with d69a4796e9347167f1a8ec166e045c53c8f362ff 🎉

PR #297, opened 2019-05-13
Send batch messages.

skipyT:


jeromegamez:
Thank you for your contribution! In the meantime, a different approach has been implemented, and will be available in the next release, probably 4.29.0.

PR #299, opened 2019-05-22
Add withSound() function Notification.php

dcenatiempo:
Currently, notifications are sent without a sound.
This PR would allow notifications to use the mobile device's default notification sound by using $message->withSound()

### DIFF
--- a/src/Firebase/Messaging/Notification.php
+++ b/src/Firebase/Messaging/Notification.php
@@ -18,6 +18,11 @@ class Notification implements \JsonSerializable
      */
     private $body;
 
+    /**
+     * @var string|null
+     */
+    private $sound;
+
     private function __construct(string $title = null, string $body = null)
     {
         $this->title = $title;
@@ -57,6 +62,14 @@ class Notification implements \JsonSerializable
         return $notification;
     }
 
+    public function withSound(): self
+    {
+        $notification = clone $this;
+        $notification->sound = 'default';
+
+        return $notification;
+    }
+
     /**
      * @return string|null
      */
@@ -78,6 +91,7 @@ class Notification implements \JsonSerializable
         return array_filter([
             'title' => $this->title,
             'body' => $this->body,
+            'sound' => $this->sound,
         ], function ($value) {
             return $value !== null;
         });


jeromegamez:
Hey @dcenatiempo, thank you for taking the time to create this PR!

The change you're proposing would only work with the legacy HTTP server protocol (see https://firebase.google.com/docs/cloud-messaging/http-server-ref) as opposed to the HTTP v1 API (see https://firebase.google.com/docs/reference/fcm/rest/v1/projects.messages).

The SDK is using the HTTP v1 API, so this will not work (and if it does, it's by coincidence and not documented). The 

With the HTTP v1 API (and therefore this SDK), to add sounds to your messages, the platform specific configuration has to be provided separately:


```php
$message = CloudMessage::withTarget('token', 'a-device-token')
    ->withNotification([
        'title' => 'Notification title',
        'body' => 'Notification body',
    ])
    ->withApnsConfig([
        'notification' => [
            'sound' => 'default',
        ]
    ])
    ->withAndroidConfig([
        'notification' => [
            'sound' => 'default'
        ]
    ]);
```

While this *is* more verbose than just using `CloudMessage::withSound()` (which, in the current state, would not allow custom sounds), it allows for more customization - for iOS, you could specify a non-default system sound or a [custom one](https://developer.apple.com/library/archive/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/SupportingNotificationsinYourApp.html#//apple_ref/doc/uid/TP40008194-CH4-SW10), and for Android the same. `default` just happens to be a special case that applies to both of them.

As both systems are different when it comes to sounds, the added flexibility justifies the slightly higher verbosity (you could also pass everything in one big array instead of using the `with*` methods.

dcenatiempo:
Thanks for you quick reply. I will make some changes and submit another PR in a bit.

jeromegamez:
@dcenatiempo Where would you make the changes? The only places where I would see them are [ApnsConfig](https://github.com/kreait/firebase-php/blob/master/src/Firebase/Messaging/ApnsConfig.php) and [AndroidConfig](https://github.com/kreait/firebase-php/blob/master/src/Firebase/Messaging/AndroidConfig.php).

Until today, I kept them as open/unspecified as they are so that the SDK doesn't restrict developers when iOS/Android parameters change and so that I as the library maintainer don't have to adapt methods when they do.

jeromegamez:
To clarify: I am totally in favor of making things easier for SDK users, so if adding methods means that, I would absolutely be for it and change methods when the specifications change. But if we add `withSound()` methods, we would also have to implement methods for every supported field on https://firebase.google.com/docs/reference/fcm/rest/v1/projects.messages for each supported target system (iOS, Android, Web)

dcenatiempo:
Nevermind, It looks like the functionality you were talking about already exists. Thanks!

jeromegamez:
As we agreed that sounds can already be added with the current feature set, I'm closing the PR now. Thanks again! 🌺

PR #311, opened 2019-08-07
Add method get users with filters

blacxi:
Hi, thanks you for the librairy, i created this PR because in my use case, i need get Firebase Users by filters on mutli attributes.
I don't know if the best way, but i added news method in the :

src/Firebase/Auth.php
src/Firebase/Auth/ApiClient.php

### DIFF
--- a/src/Firebase/Auth.php
+++ b/src/Firebase/Auth.php
@@ -107,6 +107,22 @@ class Auth
     }
 
     /**
+     * @param array $filters
+     * @return \Generator
+     * @throws Exception\AuthException
+     * @throws Exception\FirebaseException
+     */
+    public function listUsersByFilters(array $filters): \Generator
+    {
+        $response = $this->client->getUsersByFilters($filters);
+        $result = JSON::decode((string) $response->getBody(), true);
+
+        foreach ((array) ($result['users'] ?? []) as $userData) {
+            yield UserRecord::fromResponseData($userData);
+        }
+    }
+
+    /**
      * Creates a new user with the provided properties.
      *
      * @param array|Request\CreateUser $properties
--- a/src/Firebase/Auth.php
+++ b/src/Firebase/Auth.php
@@ -107,8 +107,6 @@ class Auth
     }
 
     /**
-     * @param array $filters
-     * @return \Generator
      * @throws Exception\AuthException
      * @throws Exception\FirebaseException
      */
--- a/src/Firebase/Auth/ApiClient.php
+++ b/src/Firebase/Auth/ApiClient.php
@@ -116,6 +116,20 @@ class ApiClient
     }
 
     /**
+     * @param array $filters
+     * @return ResponseInterface
+     * @throws AuthException
+     * @throws FirebaseException
+     */
+    public function getUsersByFilters(array $filters): ResponseInterface
+    {
+        return $this->request(
+            'getAccountInfo',
+            $this->transformFiltersValue($filters)
+        );
+    }
+
+    /**
      * @throws AuthException
      * @throws FirebaseException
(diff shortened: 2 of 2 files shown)


blacxi:
Hi, thanks you for the librairy, i created this PR because in my use case, i need get Firebase Users by filters on mutli attributes. 
I don't know if the best way, but i added news method in the : 
- src/Firebase/Auth.php
- src/Firebase/Auth/ApiClient.php

jeromegamez:
Hi @blacxi and thank you for considering to contribute to this repo.

Could you give me a code example on how to use this, what is filtered and how the filters work? At the moment, I honestly don't understand what the code does :/

blacxi:
Hi @jeromegamez,

I work with symfony application and use API Platform to expose all users from Firebase, for this i can use your method "listUsers()" in Auth.php.
But in my case, i added attributes (displayName...) on the users in Firebase and i need filtered the users list by one of parameters : 

```
https://local.application.com/users : 
For this case i use  "listUsers()" in Auth.php
```

```
https://local.application.com/users?displayName=testName&actived=1 : 
For this case i need use  "listUsersByFilters()" in Auth.php

    public function getListUsers(int $maxResults = self::MAX_RESULTS, int $batchSize = self::BATCH_SIZE): Generator
    {
        foreach ($this->firebase->getAuth()->listUsers($maxResults, $batchSize) as $userRecord) {
            yield $this->transformUserRecordToUser($userRecord);
        }
    }

...

    public function getListUsersByFilters(array $filters): Generator
    {
        foreach ($this->firebase->getAuth()->listUsersByFilters($filters) as $userRecord) {
            yield $this->transformUserRecordToUser($userRecord);
        }
    }
```

jeromegamez:
I just tried out your method, and especially the use case you mentioned doesn't work for me.

The SDK already has `Auth::getUserByEmail()` and `Auth::getUserByPhoneNumber()`, which the Auth API supports, but they will always return just one result because emails and phone numbers are unique.

If you got it to work, could you please give me some examples

- where other properties are supported
- where a call returns more than one result

What I tried is this:

```php
$auth = $firebase->getAuth();

$a = $auth->createUser(['displayName' => 'jerome', 'email' => uniqid().'@domain.tld']);
$b = $auth->createUser(['displayName' => 'jerome', 'email' => uniqid().'@domain.tld']);

try {
    foreach ($auth->listUsersByFilters(['displayName' => 'jerome']) as $user) {
        echo json_encode($user, JSON_PRETTY_PRINT);
    }
} catch (Throwable $e) {
    echo $e->getMessage();
} finally {
    $auth->deleteUser($a->uid);
    $auth->deleteUser($b->uid);
}
```

blacxi:
Hi, thank you for your answer.

You have right, my submission don't work. In the Firebase console, i can just use filter on "Email, PhoneNumber or UID", I think it's not possible to use filter on DisplayName or use "Like %mail_extension%" filters for get multiple users.

If you hade idea, I take any advice

Thank you

jeromegamez:
The only thing you can really do is to take the output of „listUsers“ and filter manually, unfortunately.

But I have already planned something for a future release that should make it at least more convenient.

Until then I‘ll close this PR, thank you for your contribution, I appreciate it! 🌺

blacxi:
My first idea was do this, but in my application i'm going to more than 5000 users so it's not possible to get all users in cache and after filter in.

Ping me when you will have this beautiful feature.

Thank's you for everything
See you

jeromegamez:
You don‘t need to cache all the users - the listUsers method is a generator and is really memory efficient - if you go over it with a foreach loop, it should be pretty fast.

PR #320, opened 2019-09-02
Implement link with oauth functionality

slaci:
Firebase has a [Link with OAuth credential](https://firebase.google.com/docs/reference/rest/auth#section-link-with-oauth-credential) functionality, what this PR makes possible in this lib.

The main idea is:
1. In the firebase console you enable a login provider: google, twitter, facebook etc
2. The user logs in, which is handled by a different lib
3. On successful login, it is possible to log in the user into firebase with an `id_token` or `access_token`, so the same user can use firebase stuffs

I use google login and it works for me.

Some things to note:
- The endpoint requires a `requestUri` param which is weird. I hardcoded the example `http://localhost`, because without this there is an error, but with this the login still works, so I guess it's not used. This endpoint is usable for log in too and I think this parameter makes sense there, but what I did is not for log in.
- The `postData` param can hold an `id_token` or `access_token` key. For google login both works for me, but in the examples they used the access token more, so I made that the default, but I introduced a bool param to make it possible to work with id token too
- I created constants for the `Provider` class, which I think is good. But the class is marked as internal, so PHPStorm strikes the class name when I want to use it... I don't really understand why its internal.

### DIFF
--- a/src/Firebase/Auth.php
+++ b/src/Firebase/Auth.php
@@ -11,6 +11,7 @@ use Firebase\Auth\Token\Exception\UnknownKey;
 use Generator;
 use Kreait\Firebase\Auth\ApiClient;
 use Kreait\Firebase\Auth\IdTokenVerifier;
+use Kreait\Firebase\Auth\LinkedProviderData;
 use Kreait\Firebase\Auth\UserRecord;
 use Kreait\Firebase\Exception\Auth\InvalidPassword;
 use Kreait\Firebase\Exception\Auth\RevokedIdToken;
@@ -471,6 +472,50 @@ class Auth
         $response = $this->client->unlinkProvider((string) $uid, $provider);
 
         return $this->getUserRecordFromResponse($response);
+    }
+
+    /**
+     * Logs in the user to Firebase by a provider's access token (like Google, Facebook, Twitter, etc),
+     * if the authentication provider is enabled for the project.
+     *
+     * First, you have to get a valid access token for your provider manually.
+     *
+     * @param Provider|string $provider
+     *
+     * @throws Exception\AuthException
+     * @throws Exception\FirebaseException
+     */
+    public function linkProviderThroughAccessToken($provider, string $accessToken): LinkedProviderData
+    {
+        $provider = $provider instanceof Provider ? $provider : new Provider($provider);
+        $response = $this->client->linkProviderThroughAccessToken($provider, $accessToken);
+
+        return LinkedProviderData::fromResponseData(
+            $this->getUserRecordFromResponse($response),
+            JSON::decode((string) $response->getBody(), true)
+        );
+    }
+
+    /**
+     * Logs in the user to Firebase by a provider's ID token (like Google, Facebook, Twitter, etc),
+     * if the authentication provider is enabled for the project.
+     *
+     * First, you have to get a valid ID token for your provider manually.
+     *
+     * @param Provider|string $provider
+     *
+     * @throws Exception\AuthException
+     * @throws Exception\FirebaseException
+     */
+    public function linkProviderThroughIdToken($provider, string $idToken): LinkedProviderData
+    {
+        $provider = $provider instanceof Provider ? $provider : new Provider($provider);
+        $response = $this->client->linkProviderThroughIdToken($provider, $idToken);
+
+        return LinkedProviderData::fromResponseData(
+            $this->getUserRecordFromResponse($response),
+            JSON::decode((string) $response->getBody(), true)
+        );
     }
 
     /**
--- a/src/Firebase/Auth/ApiClient.php
+++ b/src/Firebase/Auth/ApiClient.php
@@ -12,6 +12,7 @@ use Kreait\Firebase\Exception\AuthApiExceptionConverter;
 use Kreait\Firebase\Exception\AuthException;
 use Kreait\Firebase\Exception\FirebaseException;
 use Kreait\Firebase\Request;
+use Kreait\Firebase\Value\Provider;
 use Lcobucci\JWT\Token;
 use Psr\Http\Message\ResponseInterface;
 use Throwable;
@@ -304,6 +305,44 @@ class ApiClient
         return $this->request('setAccountInfo', [
             'localId' => $uid,
             'deleteProvider' => $providers,
+        ]);
+    }
+
+    /**
+     * @throws AuthException
+     * @throws FirebaseException
+     */
+    public function linkProviderThroughAccessToken(Provider $provider, string $accessToken): ResponseInterface
+    {
+        return $this->linkProvider($provider, $accessToken, 'access_token');
+    }
+
+    /**
+     * @throws AuthException
+     * @throws FirebaseException
+     */
+    public function linkProviderThroughIdToken(Provider $provider, string $idToken): ResponseInterface
+    {
+        return $this->linkProvider($provider, $idToken, 'id_token');
+    }
+
+    /**
+     * Links the given OAuth credential (e.g. Google ID token, or Facebook access token, etc) to Firebase.
+     * Basically logs in the user to Firebase, if the authentication provider is enabled for the project.
+     *
+     * @throws AuthException
+     * @throws FirebaseException
+     */
+    private function linkProvider(Provider $provider, string $token, string $tokenKeyName): ResponseInterface
+    {
+        return $this->request('https://identitytoolkit.googleapis.com/v1/accounts:signInWithIdp', [
+            'postBody' => \http_build_query([
+                $tokenKeyName => $token,
+                'providerId' => (string) $provider,
+            ]),
+            'returnSecureToken' => true,
+            'returnIdpCredential' => true,
+            'requestUri' => 'http://localhost', // this doesn't matter here, but required
         ]);
     }
 
--- a/src/Firebase/Auth/LinkedProviderData.php
+++ b/src/Firebase/Auth/LinkedProviderData.php
@@ -1,0 +1,86 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Kreait\Firebase\Auth;
+
+use Kreait\Firebase\Value\Provider;
+
+class LinkedProviderData implements \JsonSerializable
+{
+    /**
+     * @var UserRecord
+     */
+    public $userRecord;
+
+    /**
+     * @var string
+     */
+    public $federatedId;
+
+    /**
+     * @var Provider
+     */
+    public $provider;
+
+    /**
+     * The Firebase ID token.
+     *
+     * @var string
+     */
+    public $idToken;
+
+    /**
+     * @var string|null
+     */
+    public $refreshToken;
+
+    /**
+     * The provider's ID token (e.g. Google ID token).
+     *
+     * @var string|null
+     */
+    public $oauthIdToken;
+
+    /**
+     * The provider's Access token (e.g. Facebook Access token).
+     *
+     * @var string|null
+     */
+    public $oauthAccessToken;
+
+    /**
+     * @var array
+     */
+    public $rawUserInfo;
+
+    public function __construct(UserRecord $userRecord)
+    {
+        $this->userRecord = $userRecord;
+    }
+
+    public static function fromResponseData(UserRecord $userRecord, array $data): self
+    {
+        $providerData = new self($userRecord);
+
+        $providerData->federatedId = $data['federatedId'];
+        $providerData->provider = new Provider($data['providerId']);
+        $providerData->idToken = $data['idToken'];
+        $providerData->refreshToken = $data['refreshToken'];
+        $providerData->oauthIdToken = $data['oauthIdToken'] ?? null;
+        $providerData->oauthAccessToken = $data['oauthAccessToken'] ?? null;
+        $providerData->rawUserInfo = \json_decode($data['rawUserInfo'], true);
+
+        return $providerData;
+    }
+
+    public function toArray(): array
+    {
+        return \get_object_vars($this);
+    }
+
+    public function jsonSerialize()
+    {
+        return $this->toArray();
+    }
+}
--- a/src/Firebase/Exception/AuthApiExceptionConverter.php
+++ b/src/Firebase/Exception/AuthApiExceptionConverter.php
@@ -16,6 +16,7 @@ use Kreait\Firebase\Exception\Auth\InvalidPassword;
 use Kreait\Firebase\Exception\Auth\MissingPassword;
 use Kreait\Firebase\Exception\Auth\OperationNotAllowed;
 use Kreait\Firebase\Exception\Auth\PhoneNumberExists;
+use Kreait\Firebase\Exception\Auth\ProviderLinkFailed;
 use Kreait\Firebase\Exception\Auth\UserDisabled;
 use Kreait\Firebase\Exception\Auth\UserNotFound;
 use Kreait\Firebase\Exception\Auth\WeakPassword;
@@ -106,6 +107,18 @@ class AuthApiExceptionConverter
 
         if (\mb_stripos($message, 'phone_number_exists') !== false) {
             return new PhoneNumberExists('The phone number is already in use by another account.', $code, $e);
+        }
+
+        if (\mb_stripos($message, 'invalid_idp_response') !== false) {
+            return new ProviderLinkFailed('The supplied auth credential is malformed or has expired.', $code, $e);
+        }
+
+        if (\mb_stripos($message, 'invalid_id_token') !== false) {
+            return new ProviderLinkFailed('The user\'s credential is no longer valid. The user must sign in again.', $code, $e);
+        }
+
+        if (\mb_stripos($message, 'federated_user_id_already_linked') !== false) {
+            return new ProviderLinkFailed('This credential is already associated with a different user account.', $code, $e);
         }
 
         return new AuthError($message, $code, $e);
--- a/src/Firebase/Exception/Auth/ProviderLinkFailed.php
+++ b/src/Firebase/Exception/Auth/ProviderLinkFailed.php
@@ -1,0 +1,14 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Kreait\Firebase\Exception\Auth;
+
+use Kreait\Firebase\Exception\AuthException;
+use Kreait\Firebase\Exception\HasRequestAndResponse;
+use RuntimeException;
+
+final class ProviderLinkFailed extends RuntimeException implements AuthException
+{
+    use HasRequestAndResponse;
+}


jeromegamez:
Thank you @slaci for your contribution and for taking the time to create the Pull Request.

However, it is my understanding that the linking of federated Providers is something that should be done on the client-side, not the server-side: it requires user participation anyways, the client SDKs already provide the functionality to do this (
https://firebase.google.com/docs/auth/web/account-linking, https://firebase.google.com/docs/auth/android/account-linking, https://firebase.google.com/docs/auth/ios/account-linking) and this is also where the `requestUri` comes into place, because that is the URL the user gets redirected to after an account has been successfully linked.

Could you please explain why proxying everything through a server is a better approach?

None of the official Admin SDKs has this, this is only available in the Client SDKs, and it would add functionality that we can't programmatically test with Unit or Integration tests, so I'm very hesitant with adding something like this :/

Can you provide me with an example of how I can test this manually? This should probably be an example HTML page that is able to communicate with a backend, or you can also describe the steps needed to implement this.

--- 

I will add some comments to specific code after your reply and if you can convince me that this is needed in an Admin SDK (it's not , but I already want to address the notes you made:

> The endpoint requires a `requestUri` param which is weird. I hardcoded the example `http://localhost`, because without this there is an error, but with this the login still works, so I guess it's not used. This endpoint is usable for log in too and I think this parameter makes sense there, but what I did is not for log in.

If this parameter is required to make a successful request to the API, it should also be required for the `Auth::linkWithOauthCredentials()` method so that library users must explicitly specify it and know that this is a relevant value. It shouldn't be hardcoded to `http://localhost` and it should not have a default value, similar to `Auth::sendEmailVerification()` where a `$continueUrl` has to be provided.

> The `postData` param can hold an `id_token` or `access_token` key. For google login both works for me, but in the examples they used the access token more, so I made that the default, but I introduced a bool param to make it possible to work with id token too.

Being an Admin SDK, the relies on ID tokens for every operation everywhere, so we should only work with ID tokens here as well.

> I created constants for the Provider class, which I think is good. But the class is marked as internal, so PHPStorm strikes the class name when I want to use it... I don't really understand why its internal.

It's marked as internal because so far it was only used internally and was not supposed to be used directly by an implementing developer. With the added functionality it indeed makes sense to add the constants, but we should then rewrite the class so that it checks the given values if they are valid. I'll add comments to the relevant places in the code.

slaci:
> However, it is my understanding that the linking of federated Providers is something that should be done on the client-side, not the server-side: it requires user participation anyways, the client SDKs already provide the functionality to do this 

In client side it maybe do that, but when you call this REST API endpoint, then its a non-interactive one, so the user is not asked, neither redirected.

delete_fb_acc
I deleted my e-mail on the firebase console, then logged in to my app (with this modifications) and my e-mail reappered here. I didn't get redirection or any question about it.

I do the same in react-native, where the firebase login is non interactive too. They have one [lib](https://rnfirebase.io/docs/v5.x.x/auth/social-auth) for this, so the Client and Admin separation is not crystal clear for me.

The flow is this:
1. Implement google login with a different lib (like HwiOauthBundle for symfony or better just the google sdk), which is interactive, so you need a redirectUrl (requestUri?) and the user must accept the consent and whatever.
2. After the login was successful, you pass the acquired id or access token to firebase (to this API endpoint) at the server side, which links or signs in (not sure about the difference) the user to firebase silently (if the provider is enabled) and gives back a firebase token and other stuff (`UserRecord`). After this, it is possible to query the firebase db, firestore and whatever with the id token.

> Can you provide me with an example of how I can test this manually? This should probably be an example HTML page that is able to communicate with a backend, or you can also describe the steps needed to implement this.

I will have time for this at the weekend I think.

> If this parameter is required to make a successful request to the API, it should also be required for the Auth::linkWithOauthCredentials() method so that library users must explicitly specify it and know that this is a relevant value. It shouldn't be hardcoded to http://localhost and it should not have a default value, similar to Auth::sendEmailVerification() where a $continueUrl has to be provided.

If I omit the `requestUri` param then I get an error, because its missing. But if I provide one, then it doesn't do anything, because no redirection or whatsoever takes place. Maybe the client side libs use this, but not sure for what... in firebase you can't define redirect urls I think so it can't validate. The api returns an `UserRecord` data on success (except a `createdAt` field, so I made that optional), not a redirection.

> Being an Admin SDK, the relies on ID tokens for every operation everywhere, so we should only work with ID tokens here as well.

I can make the ID token the default or force it, but not sure how easy/hard to get one in FB, twitter and other PHP SDK-s. The only thing I know is that the [official google sdk](https://github.com/googleapis/google-api-php-client/blob/master/src/Google/Client.php)  `\Google_Client` class has a `getAccessToken()` method, but no accessor for the id token, so its a bit easier to use the access token here.

> It's marked as internal because so far it was only used internally and was not supposed to be used directly by an implementing developer. With the added functionality it indeed makes sense to add the constants, but we should then rewrite the class so that it checks the given values if they are valid. I'll add comments to the relevant places in the code.

I was thinking about validation, but what if firebase introduces a new provider? Then the lib must be updated to handle that, not sure if its worth it, but constants are still handy to reference these hardcoded values.

```
```

Maybe I misuse the API and the user should know about he/she will be enrolled into firebase interactively (not just from terms and services), but I haven't seen any example for that yet and it works the way I described (both in react-native and with this PR in PHP too).

jeromegamez:
Thank you for the detailed explanation, I appreciate it. I‘ll have to let it sink in a little, but if I understand you correctly, the use case is that the federated authentication isn‘t performed with one of the official client SDKs but „home made“, right?

I‘ll address the other comments you made as soon as I have some headspace... thanks again!

slaci:
>  but if I understand you correctly, the use case is that the federated authentication isn‘t performed with one of the official client SDKs but „home made“, right?

In the official [docs](https://firebase.google.com/docs/auth) they mentoin two ways for firebase auth. I think when you are talking about the client libraries, then its the first method with FirebaseUI, when the client library does the login and everything (`FirebaseUI Auth` - interactive).

My method is the second way, when you already have oauth logins with whatever libs, homemmade or official and you just pass the token to firebase which verifyes it and does the authentication non interactive if all good (`Firebase SDK Authentication`).

In the mentioned docs page there are the examples. For [Google - Web](https://firebase.google.com/docs/auth/web/google-signin) there is the [Advanced: Manual](https://firebase.google.com/docs/auth/web/google-signin#advanced:-handle-the-sign-in-flow-manually) part, which is the same what Im doing in react native and PHP. In the example you can see, they auth the user with the official google login SDK, then in the success callback they sign the user in to firebase too with the `signInWithCredential` method. This happens on the client side because this is the `web` example, but this is how it works on the serverside too. I will to create a minimal example at the weekend hopefully.

src/Firebase/Auth.php
```diff
@@ -479,4 +479,14 @@ public function unlinkProvider($uid, $provider): UserRecord
 
         return $this->getUser($uid);
     }
+
+    public function linkWithOauthCredentials($provider, string $token, bool $isAccessToken = true)
```
jeromegamez: Splitting this into two methods named

- `linkProviderThroughAccessToken($provider, string $accessToken)`
- `linkProviderThroughIdToken($provider, string $idToken)`

currently seems like a good idea to me - on the one hand, we can then remove the third bool parameter, on the other hand this makes the method names a little more "speaking" and aligns well with the `unlinkProvider()` method.

src/Firebase/Auth/ApiClient.php
```diff
@@ -297,6 +298,25 @@ public function unlinkProvider(string $uid, array $providers): ResponseInterface
         ]);
     }
 
+    /**
+     * @throws AuthException
+     * @throws FirebaseException
+     */
+    public function signInWithOauthCredentials(Provider $provider, string $token, bool $isAccessToken = true)
```
jeromegamez: In the context of the SDK, even if it's just an internal method call, I think this should be called "linkProvider*` similar to the methods in the Auth class, and probably also should be duplicated for clarity's sake (I personally don't care about code duplication here and there :))

src/Firebase/Auth.php
```diff
@@ -479,4 +479,14 @@ public function unlinkProvider($uid, $provider): UserRecord
...
+    {
+        $provider = $provider instanceof Provider ? $provider : new Provider($provider);
+        $response = $this->client->signInWithOauthCredentials($provider, $token, $isAccessToken);
+
+        $data = JSON::decode((string) $response->getBody(), true);
+
+        return UserRecord::fromResponseData($data);
```
jeromegamez: Similar to the `unlinkProvider()` method I would prefer forwarding the execution to the `getUser()` method.

```php
$uid = JSON::decode((string) $response->getBody(), true)['localId'];

return $this->getUser($uid);
```

I know this will result in an additional request, but this way we can be sure that a returned User Record always results from the same logic (my focus in the SDK, in general, is more about correctness than about performance).

src/Firebase/Exception/AuthApiExceptionConverter.php
```diff
@@ -108,6 +111,18 @@ private function convertGuzzleRequestException(RequestException $e): AuthExcepti
             return new PhoneNumberExists('The phone number is already in use by another account.', $code, $e);
         }
 
+        if (\mb_stripos($message, 'invalid_idp_response') !== false) {
```
jeromegamez: I have kept the other dedicated exceptions for backward compatibility, but for the next major release, I am planning to reduce the amount a lot. In the meantime, could you please merge the three custom exception types to one single exception? I'm not sure about the best name, but at the moment I'd go with `FederatedUserLinkingFailed` or `FailedToLinkFederatedUser`. If you have additional suggestions, please let me know ^^

jeromegamez:
> I will to create a minimal example at the weekend hopefully.

If possible I would appreciate a PR on https://github.com/jeromegamez/firebase-php-examples so that it can be tested and later reproduced and also serves as an usage-example.

I totally understand if this is much more work than you were/are ready to do, so I'll do it on my own - this would be a good training for me as well, just take longer ^^.

src/Firebase/Auth/UserMetaData.php
```diff
@@ -22,7 +22,7 @@ class UserMetaData implements \JsonSerializable
     public static function fromResponseData(array $data): self
     {
         $metadata = new self();
-        $metadata->createdAt = DT::toUTCDateTimeImmutable($data['createdAt']);
+        $metadata->createdAt = isset($data['createdAt']) ? DT::toUTCDateTimeImmutable($data['createdAt']) : null;
```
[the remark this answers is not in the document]
slaci: I fixed it, but not really needed right now because we use the `getUser` endpoint everywhere now which always returns the `createdAt` field, but I think it may be handy in the future.

src/Firebase/Auth/ApiClient.php
```diff
@@ -297,6 +298,25 @@ public function unlinkProvider(string $uid, array $providers): ResponseInterface
...
+    {
+        $tokenKeyName = $isAccessToken ? 'access_token' : 'id_token';
+
+        return $this->request('verifyAssertion', [
```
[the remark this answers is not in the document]
slaci: I changed it and it works with this url too, but the docs are really bad on this. As I understand you use the v3 (at least the base url contains v3) api and the `verifyAssertion` is the v3 endpoint, but now I changed to the v1 api. Not really sure, but both works anyway.

src/Firebase/Auth.php
slaci: I created a private function for this, because this two lines are used many times, so I think its a bit better this way.

src/Firebase/Exception/AuthApiExceptionConverter.php
slaci: I created a `LinkProviderException` because as I understand the provider is not neccesseraly federated, but maybe Im wrong. Not sure you will like the solution, but this exception handling is weird enought already so I've chosen this way.

PHPStan didn't allow me to write the assignment into the if (maybe a bug for phpstan).

slaci:
I've made some changes now:
- The things you asked for
- I tried to write some docblocks
- Some typehints (mainly return) were missing so I fixed that
- I removed the `@internal` from the Provider class, because I think no harm can be done with that little value class.

I tested the solution with both a google access token and google ID token and it works for me.

src/Firebase/Auth/ApiClient.php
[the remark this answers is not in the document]
jeromegamez: I have already on my todo list to change the URLs and API calls in general, so this will be an URL less for me to change ^^, thank you!

src/Firebase/Auth/UserMetaData.php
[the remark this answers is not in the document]
jeromegamez: I perfectly understand your intent, but if we don't need it now, I would feel more comfortable not changing it until we actually need it. This has the benefit that there's one class less to change and one test case less to cover 😅

src/Firebase/Auth.php
jeromegamez: Good choice!

src/Firebase/Exception/AuthApiExceptionConverter.php
jeromegamez: Ah, I didn't think of the "not necessarily federated"... I thought a little about it, and would like `ProviderLinkFailed` as the name of the exception (this matches the `linkProvider*` methods in the `Auth` class). You might have noticed that most exceptions don't have the `Exception` suffix in them 😓

Please also keep the message matching in the `AuthApiExceptionConverter`, please 🙏 All other exceptions are converted in there, and it removes the need for another condition/check if the static constructor of the new exception actually returned itself.

slaci:
I made the requested modifications and squashed the whole thing into a single commit. Now I think I just have to create that minimal example I promised.

slaci:
Sadly I realised that I overlooked the return data of the API. Its similar to `UserRecord` but it contains the firebase ID token too and some other data. So I propose a new value object, but not sure about the name and final form. The result partly contains UserRecord data, so I attach an UserRecord object with the result (through a separate query). The other way is to just include the `localId` and the user/dev can query the UserRecord by himself with the `getUser`.

The ID token is important, because I want to call `verifyIdToken` on it every request and this is the only way to get one (with this path). I think `verifyIdToken` has some problems when `checkIfRevoked` is true, at least it reports some brand new ID tokens (acquired through this link provider api) as revoked (depends on time, sometimes work, sometimes not). But this is a different issue, caused by this weird `$leewayInSeconds` I guess, but not totally sure, just switched this revoke check to off currently.

My example app is kinda ready, but I want to try it with one more provider, so I implemented Facebook login beside the google login, but then realized that FB requires https for the redirect url which I have some infrastructural problem atm, so needs some time. I want to know if there is any difference between the API's result between providers. There shouldn't be, but this firebase API is not really well documented, so...

slaci:
Finally I think it works. I finished the minimal example app: https://github.com/slaci/firebaseauth-example. Live demo is here: https://firebaseauth.slaci.amapa.webtown.hu

I use `linkProviderThroughIdToken` for Google and `linkProviderThroughAccessToken` for Facebook, so both works. Just they return `oauthIdToken` or `oauthAccessToken` keys, which could be merged into only one, like `oauthToken`. It is not really useful, because this is the token you give the linkProviderThrough function to authenticate, so you already have it.

The `idToken` is the most important in the `LinkedProviderData` data, with that it is possible to verify the log in.

jeromegamez:
Thank you very much for the example! I will have a deeper look in the coming week 🤞

jeromegamez:
Oh, this is good! @slaci, you're awesome! Thanks to your example application, I was able to have a good look and to better understand what's happening. Everything you did is well documented and implemented, thank you 😍.

I still am not sure that we "need" the `LinkedProviderData` class, though. I know it contains the e.g Google ID Token, Refresh Token and OAuth token of the federated account, but if the purpose is to link two user accounts (meaning: it's the same user but with different sign in methods), but we shouldn't need those fields for anything else.

I played around (combining a test script, your demo application, and updated code) and I think the `UserRecord` returned by `Auth::getUser()` contains all the information we need (again: for the purpose of linking user accounts):

```json
{
    "providerData": [
        {
            "uid": "100225589286199723571",
            "displayName": "J\u00e9r\u00f4me Gamez",
            "email": "user@gmail.com",
            "photoUrl": "https://lh3.googleusercontent.com/a-/AAuE7mD3yDp6gsOr7xlNYPjP6kVVhjjQ771wdgNu29Sh=s96-c",
            "providerId": "google.com"
        },
        {
            "uid": "+49123456789",
            "providerId": "phone",
            "phoneNumber": "+49123456789"
        },
        {
            "uid": "user@domain.tld",
            "displayName": "J\u00e9r\u00f4me Gamez",
            "email": "user@domain.tld",
            "photoUrl": "https://lh3.googleusercontent.com/a-/AAuE7mD3yDp6gsOr7xlNYPjP6kVVhjjQ771wdgNu29Sh=s96-c",
            "providerId": "password"
        }
    ],
}
```

The data might be useful for other cases, but in the "strict" scope of the Firebase SDK, I would prefer not exposing them. If we do need them for a concrete use case at a later date, we can still refer to your PR here to enable it.

If you have the time to remove the `LinkedProviderData` class and let the `linkProvider*` methods return `UserRecord`s instead, I would appreciate if you could make the changes. If you don't have the time, please let me know, then I will do it.

I am currently working on restructuring the internal processes of the SDK, so it might be that I will change some things (similar to the handling of the new Dynamic Links component), but no matter how I do it, I will make sure that you keep the full attribution for your contributed code! Thank you so much for your hard work, I appreciate it and you very much 🌺

src/Firebase/Auth/LinkedProviderData.php
```diff
@@ -0,0 +1,86 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Kreait\Firebase\Auth;
+
+use Kreait\Firebase\Value\Provider;
+
+class LinkedProviderData implements \JsonSerializable
```
jeromegamez: Please remove this class and let the methods return "just" `UserRecord`s via the `->getUser()` method (as talked about in https://github.com/kreait/firebase-php/pull/320#issuecomment-532771266)

jeromegamez:
Oh, and sorry for the back and forth on naming the methods and classes and processes, I'm not particularly good with decision on "the best" names 🙈

jeromegamez:
> The idToken is the most important in the LinkedProviderData data, with that it is possible to verify the log in.

Forget what I said/„explained“ in my last comment.

I‘m going to use your PR and build on that! It will probably take some time until a release, though 🙈

jeromegamez:
I‘m going back and forth again :). When we have linked two accounts, I don‘t think we need to verify the login itself, I believe. The identity provider (Google, Facebook, Twitter, etc.) has already verified the login, and the ID token returned by the „link*“ methods is already verified, as it has been generated by Firebase, isn‘t it?

jeromegamez:
@slaci The functionality is now in the [master branch](https://github.com/kreait/firebase-php/commits/master) and will be included in the next release, but I'd like to not make this public at the moment. You know and I know that it exists and you know how to use it, but as I'm currently revisiting my thoughts about authentication in general and restructuring the library, I might re-implement this in a different way later (the methods you provided will be available and not break for the lifetime of the 4.x release). Also, it's still a feature that I cannot easily test programmatically, and I have difficulties writing good documentation for it. 🙈 

Thank you so much for your invaluable work and this first "hidden feature" 😅, it helped me a lot understanding the "linking component" better and to organize my thoughts around the topic in general! 🌺

slaci:
Thank you!

> I played around (combining a test script, your demo application, and updated code) and I think the UserRecord returned by Auth::getUser() contains all the information we need (again: for the purpose of linking user accounts):

I think the Firebase ID token is very important and the `UserRecord` class doesn't have that. This is why I created the `LinkedProviderData`, the remaining fields are just there for completeness, they are not that useful. The Firebase ID token cannot be queried (afaik) in any way, so this is the only opportunity. The use case is: verifying the token. If the admin deletes the user in the firebase admin (or just the token expires), the app should log out the user. This is whats happening in the `refreshUser` method in my example app.
So I think its better if we keep that class 🤔

PR #348, opened 2019-10-30
Update SendMessage.php

rmsj:
The stream can only be read once. \mb_strlen((string) $body) is actually cleaning up the content of the stream.

### DIFF
--- a/src/Firebase/Messaging/Http/Request/SendMessage.php
+++ b/src/Firebase/Messaging/Http/Request/SendMessage.php
@@ -21,7 +21,7 @@ final class SendMessage implements RequestInterface
         $body = stream_for(\json_encode(['message' => $message]));
         $headers = [
             'Content-Type' => 'application/json; charset=UTF-8',
-            'Content-Length' => \mb_strlen((string) $body),
+            'Content-Length' => $body->getSize(),
         ];
 
         $this->wrappedRequest = new Request('POST', $uri, $headers, $body);


rmsj:
`<?php

require "../vendor/autoload.php";

use GuzzleHttp\Psr7\Stream;
use function GuzzleHttp\Psr7\stream_for;
use function GuzzleHttp\Psr7\uri_for;

class TestStream
{
    public static function doIt()
    {
        $resource = '{"message":{"data":{"key":"value"},"token":"dgYu9kSMsKU:APA91bFwWNCI-Kb_xzVKuxxBOB7PvjL42BaOLjfpi4FHsb82IZhBgAlgcGTqHEXEu54cqJileRca2rV1QFAvwqBwXb0R1xbZhMakX5ikkv3J9JJLqJfIyECU8PMzKwedSN7FlmluyFEe"}}';
        $data = null;
        if (is_scalar($resource)) {
            $stream = fopen('php://temp', 'r+');
            if ($resource !== '') {
                fwrite($stream, $resource);
                fseek($stream, 0);
            }
            return new Stream($stream, []);
            //print_r(fread($stream, 1000000));
        }

        return false;
    }

    public static function doItAgain()
    {
        $resource = '{"data":{"key":"value"},"token":"some_token"}';

        $uri = uri_for('https://fcm.googleapis.com/v1/projects/test/messages:send');
        $body = stream_for(\json_encode(['message' => $resource]));
        $headers = [
            'Content-Type' => 'application/json; charset=UTF-8',
            'Content-Length' => \mb_strlen((string) $body),
        ];
        print_r($headers);
        //var_dump($body->getContents());
        return $body;
    }
}

$result = TestStream::doIt();
$result2 = TestStream::doItAgain();
if ($result !== false) {
    echo "doIt: <br/>";
    print_r($result->getContents());
} else {
    echo "it failed";
}

if ($result2){
    echo "<br/><br/><br/>doItAgain: <br/>";
    print_r($result2->getContents());
}`

I'm using the laravel version of it. But the error I've found is in the changed code.
To replicate, just add this file to the public folder of a laravel project using the firebase-laravel.
Access the test.php file from the browser and see that doIt returns something. doItAgain is empty.

jeromegamez:
Nice catch! 

Because this is also done in other places in the code, would you mind applying the change also in the following locations, so that they can be attributed to you as well?

- https://github.com/kreait/firebase-php/blob/master/src/Firebase/DynamicLink/CreateDynamicLink/ApiRequest.php#L25
- https://github.com/kreait/firebase-php/blob/master/src/Firebase/DynamicLink/ShortenLongDynamicLink/ApiRequest.php#L25
- https://github.com/kreait/firebase-php/blob/master/src/Firebase/Messaging/Http/Request/SendMessage.php#L24
- https://github.com/kreait/firebase-php/blob/master/src/Firebase/Messaging/Http/Request/ValidateMessage.php#L24

No problem, if you don't want to, then I'll merge this PR and add an additional commit.

jeromegamez:
Merged with https://github.com/kreait/firebase-php/commit/fe8c1f3b4fa2e3269dc1435b3c27ba5a7e9b9259

I made the changes in the other files as well and attributed them to you. Thanks again! 🌺

PR #356, opened 2019-11-27
Adds the full reset password procedure

joemugen:
# Adds the full reset password procedure

## What?

- `src/Firebase/Auth/ApiClient.php`
- `src/Firebase/Auth.php`

## Why?

I had the need to verify if the oobCode was valid before processing the request and updating a password.
Then I had to reset the password with the given oobCode.

On the official Firebase documentation and SDK, I found these methods were existing (verifyPasswordResetCode and confirmPasswordReset). I thought it was a good idea to implement it in this package too.

## How ?

It modifies the `src/Firebase/Auth/ApiClient.php` file by adding the `verifyPasswordResetCode` and `confirmPasswordReset` based on the official Firebase documentation (https://firebase.google.com/docs/reference/rest/auth#section-verify-password-reset-code and https://firebase.google.com/docs/reference/rest/auth#section-confirm-reset-password)

**About the password reset code verification**

If it fails, it will throw an Exception with one of these errors:

- OPERATION_NOT_ALLOWED: Password sign-in is disabled for this project.
- EXPIRED_OOB_CODE: The action code has expired.
- INVALID_OOB_CODE: The action code is invalid. This can happen if the code is malformed, expired, or has already been used.

**And for the password reset confirmation**
It it fails, it will throw an Exception with one of these errors:

- OPERATION_NOT_ALLOWED: Password sign-in is disabled for this project.
- EXPIRED_OOB_CODE: The action code has expired.
- INVALID_OOB_CODE: The action code is invalid. This can happen if the code is malformed, expired, or has already been used.
- USER_DISABLED: The user account has been disabled by an administrator

I also added the method `verifyPasswordResetCode` and `confirmPasswordReset` in `src/Firebase/Auth.php`. It just call the previous methods and returns the response body.

I know both are kinda the same but there is 2 methods in the official SDK. Anyway we can merge both method in `ApiClient` and make `$newPassword` nullable. Then we build the array with or without the `newPassword` key.

Exemple of my custom rule in Laravel if you are interested by a use case:
```php
<?php

namespace App\Rules;

use Illuminate\Contracts\Validation\Rule;
use Kreait\Laravel\Firebase\Facades\FirebaseAuth;

class isValidFirebaseOobCode implements Rule
{
    public const FIREBASE_VALID_REQUEST_TYPE = 'PASSWORD_RESET';

    public function __construct()
    {
        //
    }
    
    public function passes($attribute, $value): bool
    {
        try {
            // If the verification fails, it will throw an exception
            $response = FirebaseAuth::verifyPasswordResetCode($value);


            return $response['requestType'] === self::FIREBASE_VALID_REQUEST_TYPE;
        } catch (\Exception $exception) {
            return false;
        }
    }

    public function message()
    {
        return __('The password reset link is invalid or no longer valid. Please ask for a new reset password link.');
    }
}

```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,5 @@
 /composer.lock
 /phpunit.xml
 /google-service-account.json
+
+.idea
--- a/.gitignore
+++ b/.gitignore
@@ -11,5 +11,3 @@
 /composer.lock
 /phpunit.xml
 /google-service-account.json
-
-.idea
--- a/src/Firebase/Auth.php
+++ b/src/Firebase/Auth.php
@@ -471,7 +471,7 @@ class Auth
      * - USER_DISABLED: The user account has been disabled by an administrator.
      *
      * @param string $oobCode the email action code sent to the user's email for resetting the password
-     * @param string $newPassword the user's new password
+     * @param mixed $newPassword the user's new password
      *
      * @throws Exception\AuthException
      * @throws Exception\FirebaseException
@@ -480,8 +480,10 @@ class Auth
      *
      * @see https://firebase.google.com/docs/reference/rest/auth#section-confirm-reset-password
      */
-    public function confirmPasswordReset(string $oobCode, string $newPassword)
+    public function confirmPasswordReset(string $oobCode, $newPassword)
     {
+        $newPassword = $newPassword instanceof ClearTextPassword ? $newPassword : new ClearTextPassword($$newPassword);
+
         $response = $this->client->confirmPasswordReset($oobCode, $newPassword);
 
         return JSON::decode((string) $response->getBody(), true);
--- a/src/Firebase/Auth.php
+++ b/src/Firebase/Auth.php
@@ -445,7 +445,7 @@ class Auth
      * - EXPIRED_OOB_CODE: The action code has expired.
      * - INVALID_OOB_CODE: The action code is invalid. This can happen if the code is malformed, expired, or has already been used.
      *
-     * @param string $oobCode
+     * @param string $oobCode The email action code sent to the user's email for resetting the password.
      * @return mixed
      * @throws Exception\AuthException
      * @throws Exception\FirebaseException
@@ -456,6 +456,29 @@ class Auth
         $response = $this->client->verifyPasswordResetCode($oobCode);
 
         return JSON::decode((string) $response->getBody(), true);
+    }
+
+    /**
+     * Confirm the reset password
+     *
+     * 4 errors code possible:
+     * - OPERATION_NOT_ALLOWED: Password sign-in is disabled for this project.
+     * - EXPIRED_OOB_CODE: The action code has expired.
+     * - INVALID_OOB_CODE: The action code is invalid. This can happen if the code is malformed, expired, or has already been used.
+     * - USER_DISABLED: The user account has been disabled by an administrator.
+     *
+     * @param string $oobCode The email action code sent to the user's email for resetting the password.
+     * @param string $newPassword The user's new password.
+     * @return mixed
+     * @throws Exception\AuthException
+     * @throws Exception\FirebaseException
+     * @see https://firebase.google.com/docs/reference/rest/auth#section-confirm-reset-password
+     */
+    public function confirmPasswordReset(string $oobCode, string $newPassword)
+    {
+        $response = $this->client->confirmPasswordReset($oobCode, $newPassword);
+
+        return JSON::decode((string)$response->getBody(), true);
     }
 
     /**
--- a/src/Firebase/Auth.php
+++ b/src/Firebase/Auth.php
@@ -72,9 +72,9 @@ class Auth
     {
         $uid = $uid instanceof Uid ? $uid : new Uid($uid);
 
-        $response = $this->client->getAccountInfo((string) $uid);
+        $response = $this->client->getAccountInfo((string)$uid);
 
-        $data = JSON::decode((string) $response->getBody(), true);
+        $data = JSON::decode((string)$response->getBody(), true);
 
         if (empty($data['users'][0])) {
             throw new UserNotFound("No user with uid '{$uid}' found.");
@@ -84,10 +84,10 @@ class Auth
     }
 
     /**
-     * @throws Exception\AuthException
+     * @return Generator|UserRecord[]
      * @throws Exception\FirebaseException
      *
-     * @return Generator|UserRecord[]
+     * @throws Exception\AuthException
      */
     public function listUsers(int $maxResults = 1000, int $batchSize = 1000): Generator
     {
@@ -96,9 +96,9 @@ class Auth
 
         do {
             $response = $this->client->downloadAccount($batchSize, $pageToken);
-            $result = JSON::decode((string) $response->getBody(), true);
+            $result = JSON::decode((string)$response->getBody(), true);
 
-            foreach ((array) ($result['users'] ?? []) as $userData) {
+            foreach ((array)($result['users'] ?? []) as $userData) {
                 yield UserRecord::fromResponseData($userData);
 
                 if (++$count === $maxResults) {
@@ -177,9 +177,9 @@ class Auth
     {
         $email = $email instanceof Email ? $email : new Email($email);
 
-        $response = $this->client->getUserByEmail((string) $email);
+        $response = $this->client->getUserByEmail((string)$email);
 
-        $data = JSON::decode((string) $response->getBody(), true);
+        $data = JSON::decode((string)$response->getBody(), true);
 
         if (empty($data['users'][0])) {
             throw new UserNotFound("No user with email '{$email}' found.");
@@ -198,9 +198,9 @@ class Auth
     {
         $phoneNumber = $phoneNumber instanceof PhoneNumber ? $phoneNumber : new PhoneNumber($phoneNumber);
 
-        $response = $this->client->getUserByPhoneNumber((string) $phoneNumber);
+        $response = $this->client->getUserByPhoneNumber((string)$phoneNumber);
 
-        $data = JSON::decode((string) $response->getBody(), true);
+        $data = JSON::decode((string)$response->getBody(), true);
 
         if (empty($data['users'][0])) {
             throw new UserNotFound("No user with phone number '{$phoneNumber}' found.");
@@ -276,7 +276,7 @@ class Auth
         $uid = $uid instanceof Uid ? $uid : new Uid($uid);
 
         try {
-            $this->client->deleteUser((string) $uid);
+            $this->client->deleteUser((string)$uid);
         } catch (UserNotFound $e) {
             throw new UserNotFound("No user with uid '{$uid}' found.");
         }
@@ -292,14 +292,14 @@ class Auth
     public function sendEmailVerification($uid, $continueUrl = null, string $locale = null)
     {
         if ($continueUrl !== null) {
-            $continueUrl = (string) $continueUrl;
+            $continueUrl = (string)$continueUrl;
         }
 
         $response = $this->client->exchangeCustomTokenForIdAndRefreshToken(
             $this->createCustomToken($uid)
         );
 
-        $idToken = JSON::decode((string) $response->getBody(), true)['idToken'];
+        $idToken = JSON::decode((string)$response->getBody(), true)['idToken'];
 
         $this->client->sendEmailVerification($idToken, $continueUrl, $locale);
     }
@@ -314,12 +314,12 @@ class Auth
     public function sendPasswordResetEmail($email, $continueUrl = null, string $locale = null)
     {
         if ($continueUrl !== null) {
-            $continueUrl = (string) $continueUrl;
+            $continueUrl = (string)$continueUrl;
         }
 
-        $email = $email instanceof Email ? $email : new Email((string) $email);
+        $email = $email instanceof Email ? $email : new Email((string)$email);
 
-        $this->client->sendPasswordResetEmail((string) $email, (string) $continueUrl, $locale);
+        $this->client->sendPasswordResetEmail((string)$email, (string)$continueUrl, $locale);
     }
 
     /**
@@ -374,8 +374,11 @@ class Auth
      * @throws Exception\FirebaseException
      * @throws Exception\AuthException
      */
-    public function verifyIdToken($idToken, bool $checkIfRevoked = false, /* @deprecated */ bool $allowTimeInconsistencies = null): Token
-    {
+    public function verifyIdToken(
+        $idToken,
+        bool $checkIfRevoked = false,
+        /* @deprecated */ bool $allowTimeInconsistencies = null
+    ): Token {
         // @codeCoverageIgnoreStart
         if (\is_bool($allowTimeInconsistencies)) {
             // @see https://github.com/firebase/firebase-admin-dotnet/pull/29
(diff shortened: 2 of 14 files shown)


jeromegamez:
Hey @joemugen! I just tested this manually and it works as promised 🎉.

A couple of notes:

- Could you please run `vendor/bin/php-cs-fixer fix` once and push the changes, this will reduce the changeset quite a bit
- We should make use of the [`Kreait\Firebase\Exception\AuthApiExceptionConverter`](https://github.com/kreait/firebase-php/blob/master/src/Firebase/Exception/AuthApiExceptionConverter.php) to handle the specific errors - `USER_DISABLED` and `OPERATION_NOT_ALLOWED` are already handled, so two new cases for `EXPIRED_OOB_CODE` and `INVALID_OOB_CODE` with two new exceptions similar to [`Kreait\Firebase\Exception\Auth\OperationNotAllowed`](https://github.com/kreait/firebase-php/blob/master/src/Firebase/Exception/Auth/OperationNotAllowed.php) and as exception message the descriptions you used in your PR description.
- As these are new methods, I'd like for them to have no return value (void) in the `Auth` class: either the operation is successful or it throws one of the specific exceptions. The ApiClient can continue to return the responses, even if it's not used.

It would be great if you would like to implement those changes - if not, no problem, then I'll do it and make sure that this contribution remains attributed to you.

Thank you for taking the time and for putting work into making the SDK better! 🌺

joemugen:
You're welcome, thanks to you for the hard work!

I'm close to have everything done except this point:
> * As these are new methods, I'd like for them to have no return value (void) in the `Auth` class: either the operation is successful or it throws one of the specific exceptions. The ApiClient can continue to return the responses, even if it's not used.

I understand your concern but here I need to grab some value, in `Auth::verifyPasswordResetCode` I need the `requestType` and in the `Auth::confirmPasswordReset` I need the email in the response.

how would you like me to do? 

Using a similar approach of `getUserRecordFromResponse` but storing the response in Auth private properties will do? 

The _why_ I'd like these in Auth it's because of the FirebaseAuth facade in Laravel :)

.gitignore
```diff
@@ -11,3 +11,5 @@
 /composer.lock
 /phpunit.xml
 /google-service-account.json
+
+.idea
```
jeromegamez: I usually recommend having this in a global .gitignore file (together with OS- and other editor-specific files), so that the local file is really just project related 🙈

src/Firebase/Auth.php
```diff
@@ -437,6 +437,56 @@ public function verifyPassword($email, $password): UserRecord
...
+     * @param string $newPassword the user's new password
+     *
+     * @throws Exception\AuthException
+     * @throws Exception\FirebaseException
+     *
+     * @return mixed
+     *
+     * @see https://firebase.google.com/docs/reference/rest/auth#section-confirm-reset-password
+     */
+    public function confirmPasswordReset(string $oobCode, string $newPassword)
```
jeromegamez: For consistency, I would handle the password similar to the `verifyPassword()` method:

```php
/**
 * ...
 * @param Email|string $email
 * @param ClearTextPassword|string $newPassword
 *
 * @throws InvalidArgumentException if the given password is invalid
 * ...
 */
public function confirmPasswordReset(string $oobCode, $newPassword)
{
    $newPassword = $newPassword instanceof ClearTextPassword ? $newPassword : new ClearTextPassword($$newPassword);

    $response = $this->client->confirmPasswordReset($oobCode, (string) $newPassword);
    // ...
}
```

src/Firebase/Auth/ApiClient.php
```diff
@@ -74,13 +74,13 @@ public function updateUser(Request\UpdateUser $request): ResponseInterface
     }
 
     /**
+     * @throws AuthException
```
jeromegamez: Could you please revert the changes in the PHPDocs except for the new methods? 🙈 (I make this emoji because, on the one hand, I know that this is just a minor/non-relevant thing, but on the other hand, it's important to me 😅)

src/Firebase/Database.php
```diff
@@ -76,10 +76,7 @@ public function getReferenceFromUrl($uri): Reference
         $uri = $uri instanceof UriInterface ? $uri : new Uri($uri);
 
         if (($givenHost = $uri->getHost()) !== ($dbHost = $this->uri->getHost())) {
-            throw new InvalidArgumentException(\sprintf(
-                'The given URI\'s host "%s" is not covered by the database for the host "%s".',
-                $givenHost, $dbHost
-            ));
+            throw new InvalidArgumentException(\sprintf('The given URI\'s host "%s" is not covered by the database for the host "%s".', $givenHost, $dbHost));
```
jeromegamez: Could you please revert the changes in this (unrelated) class? 🙈

src/Firebase/Database/Reference/Validator.php
```diff
@@ -35,20 +35,14 @@ private function validateDepth(string $path)
         $depth = \mb_substr_count($path, '/') + 1;
 
         if ($depth > self::MAX_DEPTH) {
-            throw new InvalidArgumentException(\sprintf(
```
jeromegamez: Could you please revert the changes in this (unrelated) class? 🙈

src/Firebase/Exception/AuthApiExceptionConverter.php
```diff
@@ -121,6 +123,22 @@ private function convertGuzzleRequestException(RequestException $e): AuthExcepti
             return new ProviderLinkFailed('This credential is already associated with a different user account.', $code, $e);
         }
 
+        if (\mb_stripos($message, 'USER_DISABLED') !== false) {
+            return new UserDisabled('The user account has been disabled by an administrator.', $code, $e);
+        }
+
+        if (\mb_stripos($message, 'OPERATION_NOT_ALLOWED') !== false) {
+            return new OperationNotAllowed('Operation not allowed.', $code, $e);
+        }
```
jeromegamez: USER_DISABLED and OPERATION_NOT_ALLOWED are already handled above

src/Firebase/Exception/AuthApiExceptionConverter.php
```diff
@@ -121,6 +123,22 @@ private function convertGuzzleRequestException(RequestException $e): AuthExcepti
...
+
+        if (\mb_stripos($message, 'EXPIRED_OOB_CODE') !== false) {
```
jeromegamez:
```suggestion
        if (\mb_stripos($message, 'expired_oob_code') !== false) {
```

`mb_stripos()` is case-insensitive and all other checks work with lowercased string. I implemented the checks in this way to be prepared just in case™ 😅

src/Firebase/Exception/AuthApiExceptionConverter.php
```diff
@@ -121,6 +123,22 @@ private function convertGuzzleRequestException(RequestException $e): AuthExcepti
...
+            return new ExpiredOobCode('The action code has expired.', $code, $e);
+        }
+
+        if (\mb_stripos($message, 'INVALID_OOB_CODE') !== false) {
```
jeromegamez:
```suggestion
        if (\mb_stripos($message, 'invalid_oob_code') !== false) {
```

`mb_stripos()` is case-insensitive and all other checks work with lowercased string. I implemented the checks in this way to be prepared just in case™ 😅

src/Firebase/Messaging.php
```diff
@@ -212,9 +212,7 @@ private function makeMessage($message): Message
         }
 
         if (!\is_array($message)) {
-            throw new InvalidArgumentException(
```
jeromegamez: Could you please revert the changes in this (unrelated) class? 🙈

src/Firebase/Messaging/CloudMessage.php
```diff
@@ -62,10 +62,7 @@ public static function fromArray(array $data)
         $new = new static();
 
         if (\count(\array_intersect(\array_keys($data), MessageTarget::TYPES)) > 1) {
-            throw new InvalidArgument(
```
jeromegamez: Could you please revert the changes in this (unrelated) class? 🙈

src/Firebase/RemoteConfig/TagColor.php
```diff
@@ -35,11 +35,7 @@ public function __construct(string $value)
         $value = \mb_strtoupper($value);
 
         if (!\in_array($value, self::VALID_COLORS, true)) {
-            throw new InvalidArgumentException(
```
jeromegamez: Could you please revert the changes in this (unrelated) class? 🙈

src/Firebase/ServiceAccount.php
```diff
@@ -142,13 +142,7 @@ public static function fromArray(array $config): self
         }
 
         if (!empty($missingFields)) {
-            throw new InvalidArgumentException(
```
jeromegamez: Could you please revert the changes in this (unrelated) class? 🙈

src/Firebase/ServiceAccount/Discovery/FromEnvironmentVariable.php
```diff
@@ -41,9 +41,7 @@ public function __invoke(): ServiceAccount
         try {
             return (new FromPath($path))();
         } catch (ServiceAccountDiscoveryFailed $e) {
-            throw new ServiceAccountDiscoveryFailed(
```
jeromegamez: Could you please revert the changes in this (unrelated) class? 🙈

src/Firebase/Util/JSON.php
```diff
@@ -31,8 +31,7 @@ public static function encode($value, $options = null, $depth = null): string
 
         $json = \json_encode($value, $options, $depth);
         if (\json_last_error() !== \JSON_ERROR_NONE) {
-            throw new InvalidArgumentException(
```
jeromegamez: Could you please revert the changes in this (unrelated) class? 🙈

jeromegamez:
Concerning the response contents...

I don't think that you need the request type as it's always `PASSWORD_RESET` and it's the result of a call to `verifyPasswordResetCode()`, so I would have assumed that in the Laravel rule this should be enough:

```php
public function passes($attribute, $value): bool
{
    try {
        // If the verification fails, it will throw an exception
        FirebaseAuth::verifyPasswordResetCode($value);

        return true;
    } catch (InvalidOobCode|ExpiredOobCode $e) {
        return false;
    }
}
```

As for the `confirmPasswordReset()` method, I would argue that its purpose is to update the password for the given OOB code, explicitly without involving the email address - otherwise the `changeUserPassword()` would have been enough and now we would have a mapping between `$newPassword` and the returned email that could be hijacked by a malicious player.

Of course, I can't (and don't want to) tell you how your use case works and what you should be doing, but as I only think in the scope of this SDK, I would still favor not having return values for the two new methods.

However! If you say: "but I want/need them", then I would agree to keep them in the 4.x releases (but not document the returns), with the caveat that they will certainly be removed in the next major (5.0) release ¯\\_(ツ)_/¯

joemugen:
> Concerning the response contents...
> 
> I don't think that you need the request type as it's always `PASSWORD_RESET` and it's the result of a call to `verifyPasswordResetCode()`, so I would have assumed that in the Laravel rule this should be enough:
> 
> ```
> public function passes($attribute, $value): bool
> {
>     try {
>         // If the verification fails, it will throw an exception
>         FirebaseAuth::verifyPasswordResetCode($value);
> 
>         return true;
>     } catch (InvalidOobCode|ExpiredOobCode $e) {
>         return false;
>     }
> }
> ```
> 
> As for the `confirmPasswordReset()` method, I would argue that its purpose is to update the password for the given OOB code, explicitly without involving the email address - otherwise the `changeUserPassword()` would have been enough and now we would have a mapping between `$newPassword` and the returned email that could be hijacked by a malicious player.
> 
> Of course, I can't (and don't want to) tell you how your use case works and what you should be doing, but as I only think in the scope of this SDK, I would still favor not having return values for the two new methods.
> 
> However! If you say: "but I want/need them", then I would agree to keep them in the 4.x releases (but not document the returns), with the caveat that they will certainly be removed in the next major (5.0) release ¯\_(ツ)_/¯

I totally agree about the rule use case and the `verifyPasswordResetCode`

But for the other use case, I think I have to give you some details.

- Using the reset password (and not update) because the user is not logged in and lost his crendentials.
- Once the reset password is done, I have no idea of his email (because I can't modify the Firebase email URL with %EMAIL% sadly D:) and I need the email to get the user in order to invalid his previously issued tokens.
- The only way is to grab back the email in the `confirmPasswordReset`

If you prefer to have only void response in Auth, It will force the use of the API client through the Auth class.
And instead of that (using Laravel facade):
```php
<?php
/* [...] */
$response = FirebaseAuth::confirmPasswordReset($request->get('oobCode'), $request->get('password'));

$firebaseUser = FirebaseAuth::getUserByEmail($response['email']);
```

We'll have to do something like this:
```php
<?php
$response = FirebaseAuth::getApiClient()->confirmPasswordReset($request->get('oobCode'), $request->get('password'));

$email = json_decode($response->getBody(), true)['email'];
        
$firebaseUser = FirebaseAuth::getUserByEmail($email);
```

I'm OK doing your way, but in this case there is no point keeping `confirmPasswordReset` in `Auth` and it can be confusing.

.gitignore
joemugen: You're right. I should do that :o

src/Firebase/Database.php
joemugen: I think the php-cs fixer does _this fix_ :p 
I'll fix that too :)

jeromegamez: I was surprised as well because it doesn't on my machine™ 😅

jeromegamez:
Thanks for detailing the use case, now it absolutely makes sense to me! What would you think about changing the method like this?

```php
public function confirmPasswordReset(string $oobCode, string $newPassword, bool $invalidateSessions = null)
{
    
    $response = $this->client->confirmPasswordReset($oobCode, $newPassword);
    $email = JSON::decode((string) $response->getBody(), true)['email'] ?? null;

    if ($invalidateSessions && $email) {
        $this->revokeRefreshTokens($this->getUserByEmail($email)->uid);
    }
}
```

Would this be a way that we could both be happy with? 😅

Perhaps we could also think about a better name for the method then, perhaps `resetPasswordWithActionCode()`? What do you think?

joemugen:
I like this approach. The only cons I see, we're not sticking to the official SDKs and the official documentation.
`confirmPasswordReset` is clearly not obvious but used in the JS SDK https://firebase.google.com/docs/reference/js/firebase.auth.Auth#methods

I like both, tell me what your prefer :)

jeromegamez:
I prefer mine, of course! 😂

But in all seriousness: I diverged from the official SDKs in quite some places already, and am trying not to do it in a certain way just because the official SDKs are doing something, when I think that it makes more sense. I think that‘s the beauty of being unofficial 🤭

joemugen:
> I prefer mine, of course! 
> 
> But in all seriousness: I diverged from the official SDKs in quite some places already, and am trying not to do it in a certain way just because the official SDKs are doing something, when I think that it makes more sense. I think that‘s the beauty of being unofficial 🤭

OK no problem at all with that. I will do the modification ASAP

joemugen:
Looks like you'll have to run php-cs if you want to be happy with Travis :p

https://travis-ci.org/kreait/firebase-php/jobs/618216579?utm_medium=notification&utm_source=github_status

jeromegamez:
I used your contribution as the starting point for #358 and hope to have it in the next release soon! I'd like for you to keep an eye on the PR as I will apply some changes that we already discussed (return values or lack thereof) and probably in the naming of the methods and classes as well (I'm not sure yet if I want to expose the "OOB" terminology - I just hat to look the acronym up to know what it actually meant 😅)

Thanks again for the work and time you've already put into this! 🌺

Issue #359, opened 2019-12-09
generateSignInWithEmailLink(...) in Admin Auth DSK


mkilmanas:
**Is your feature request related to a problem? Please describe.**
We are implementing Firebase authentication on our project, which uses Javascript Client SDK for direct user interaction, but we want to use PHP Admin SDK for generating and sending the login links by email directly to the user [(Basically this thing)](https://firebase.google.com/docs/auth/admin/email-action-links#generate_email_link_for_sign-in) But I see this particular method is not (yet?) implemented in this lib.

**Describe the solution you'd like**
I would love to see the `generateSignInWithEmailLink(...)` method added to the Auth SDK

**Describe alternatives you've considered**
Using the Admin SDK from another language?... but that seems like a huge overkill to add a new language to the "zoo" just for the sake of this single call.

**Additional context**
[Here](https://github.com/kreait/firebase-php/issues/312#issuecomment-520166635) it was mentioned that it is likely to be added in the future release.
With some guidance on where to start and how to properly implement this I don't mind getting my hands dirty and I could contribute in the form of PR(s).
But I'm rather fresh to Firebase terminology and I'm not quite fully comprehending the structure/architecture of this library, so might need quite some help.

jeromegamez:
Thanks for letting me know/reminding me! I‘ve had a Look at the other SDKs and the internals have changed a little since I last touched the component, so it will take me a bit to apply this to the library - but know that this is now on the top of my todo list.

mkilmanas:
Glad to hear! Let me know if I can help in any way. Looking forward to this

jeromegamez:
@mkilmanas I just created a draft PR at #360 which should include the basic functionality, would you like to give it a try and see if it works for you so far? There is still a lot missing, but I believe I'm onto something ^^.

mkilmanas:
@jeromegamez looks great! I've tried it out and seems to work nicely. 
All but one thing - is it currently respecting the ActionCodeSettings parameter? From what I found in the [Firebase Docs](https://firebase.google.com/docs/auth/admin/email-action-links#initialize_actioncodesettings) the 'url' parameter should mean the 'continueUrl', i.e. where to redirect upon verification (or did I misinterpret that?), but currently I get "Error encountered / Continue URL is required for email sign-in!". It only works if I manually append `'&continueUrl=' . urlencode($continueUrl);` at the end of the retrieved url. Is that expected? Or is that To-Do?
[...]

PR #360, opened 2019-12-10
Rework the action link handling for better resilience and flexibility

jeromegamez:
Implementation of the feature request in #359, including a rework of the handling in general

### How to use

Add `"kreait/firebase-php": "dev-action-links"` to your composer.json file

```php
use Kreait\Firebase\Auth\CreateActionLink\FailedToCreateActionLink;
use Kreait\Firebase\Auth\CreateActionLink\FailedToSendActionLink;

$email = 'user@domain.tld';
$actionCodeSettings = null; // Optional: null, array, or instance of ActionCodeSettings (see below)
$locale = null; // Optional: string|null

$auth = $factory->createAuth();

// Get the action links to use in custom emails
try {
    echo $auth->getPasswordResetLink($email, $actionCodeSettings).PHP_EOL;
    echo $auth->getEmailVerificationLink($email, $actionCodeSettings).PHP_EOL;
    echo $auth->getSignInWithEmailLink($email, $actionCodeSettings).PHP_EOL;
} catch (FailedToCreateActionLink $e) {
    // ...
}

// Let firebase send the emails for us

try {
    $auth->sendPasswordResetLink($email, $actionCodeSettings, 
$locale);
    $auth->sendEmailVerificationLink($email, $actionCodeSettings, $locale);
    $auth->sendSignInWithEmailLink($email, $actionCodeSettings, $locale);
} catch (FailedToSendActionLink $e) {
    // ...
}
```

#### Action code settings

Can be provided as an array, the following keys are supported:

- `continueUrl`/`url`
- `canHandleCodeInApp`
- `dynamicLinkDomain`
- `androidPackageName`
- `androidMinimumVersion`
- `androidInstallApp`
- `iOSBundleId`

### Todo

- [x] Add support for action code settings (the class is already there, but neither worked out nor used 🤷‍♂)
- [x] Validate action code settings
- [x] Add/Modify tests
- [ ] Add/Modify documentation
- [x] Let the existing methods use the new methods
- [x] Add more convenience methods OR deprecate existing convenience methods 🤔 
- [x] `Auth::sendEmailVerification()`/`Auth::sendEmailActionLink('VERIFY_EMAIL')` is currently broken, apparently the email is not enough, the API endpoint needs an ID token as well (see https://github.com/firebase/firebase-js-sdk/issues/1958)

:octocat:


jeromegamez:
@mkilmanas I think the feature is ready (please note the updated usage in the PR description). The only missing part is the hardest: update the documentation 😅. Please let me know if something doesn't work out in case you test this again 🙏

mkilmanas:
@jeromegamez - I've just tried the updated version and it seems to be running smoothly 👍 My main focus was on `getSignInWithEmailLink(...)` but I've also tried the others at least once as well.
Continue URL is working via actionCodeSettings fine 👍 
A couple of things to note (may or may not need changing):

- [Official Docs](https://firebase.google.com/docs/auth/admin/email-action-links#initialize_actioncodesettings) say that parameter name for ActionCodeSettings is called `url` and then interpreted/used as `continueUrl`, however in your implementation the setting has to be named 'continueUrl' in order for it to work. Not a huge issue, but may be misleading to others as it deviates from the docs;
 - Class names for ActionCodeSettings implementations are a little strange when used without full namespace:
```php
$actionCodeSettings = new Raw(['foo' => 'bar']); // Raw who?..
```
And I think most people will use it this way, with namespace being at the top in the `use` section
- I love how you made named methods for each use case instead of public generic one 👍 But in your usage example above you forgot to remove the first parameter from the `sendXYZ()` calls
Thanks for the great work and a superb quick turnaround - greatly appreciated 🍻

jeromegamez:
Thank you for your feedback! I renamed the classes to `RawActionCodeSettings` and `ValidatedActionCodeSettings` to make the usage clearer (I always switch in my thinking of preferring explicit names and "composed" names from namespace + class name, where I then say "You could always do `use Long\Namespace\Short\Class as ShortClass`, but here I obviously want to please you 😂 )

Concerning the action code settings, I added `url` as a synonym to `continueUrl` to cover both cases - but the inconsistencies to the official documentation are still there, e.g. where I support `iOSBundleId`, the Admin Node SDK wants developers to use `{"iOS": {"bundleId": "xxx"}}`. I have some components that are slightly different in parameter and methods names than the official SDKs, but I also have my "own" documentation, so I hope developers look at that first instead of trying to apply the official docs 1:1 to this SDK. What do you think about it?

jeromegamez:
@mkilmanas I just update the docs in https://github.com/kreait/firebase-php/blob/action-links/docs/user-management.rst#using-email-action-codes and I think the feature is ready now. You have until the end of the current build to object 😂 

No, seriously: if there's something missing or needing a change, I think we can do it separately - at the moment I don't think there should be a no-go left.

Thank you so much for your cooperation and for testing this (and of course for triggering me in the first place) :hibi

mkilmanas:
No, thank you for your rapid implementation and great communication 👍

PR #376, opened 2020-01-31
Multicast send reports improvements

kprigel:
In order to use the multicast send report in order to determine invalid tokens, added a method everything() to the MulticastSendReport class, and added getTarget() to the CloudMessage class.  This allows one to iterate $multicastSendReport->everything()->getItems() and use an index to determine the target of the failed messages from the $messages array: $messages[$idx]->getTarget()->value().

### DIFF
--- a/src/Firebase/Messaging/CloudMessage.php
+++ b/src/Firebase/Messaging/CloudMessage.php
@@ -209,6 +209,11 @@ class CloudMessage implements Message
         return $this->target ? true : false;
     }
 
+    public function  getTarget()
+    {
+        return $this->target;
+    }
+    
     public function jsonSerialize()
     {
         $data = [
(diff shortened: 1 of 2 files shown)


jeromegamez:
Could you elaborate on what particular issue this solves? Unless I'm missing something, `$multicastSendReport->everything()->getItems()` is effectively identical to `$multicastSendReport->getItems()`.

And you don't need `CloudMessage::getTarget()` because no instance of CloudMessage will ever be in a send report.

So, to fulfill what you're describing, you can do already:

```php
$multicastSendReport = $messaging->sendMulticast($message, $targets);

foreach ($multicastSendReport->getItems() as $report) {
    if ($report->isFailure()) {
        echo "Message to {$report->target()->type()} '{$report->target()->value()}' failed: \n";
        print_r($report->result());
    } elseif ($report->isSuccess()) {
        echo "Message to {$report->target()->type()} '{$report->target()->value()}' succeeded: \n";
        print_r($report->result());
    } else {
        echo "This will never be shown\n";
    }
}
```

Am I missing something?

kprigel:
You are correct, your method works.

PR #378, opened 2020-02-02
Apply leeway to all checked timestamps

### LINKED ISSUE #377: Removing leeway configuration to avoid waiting
Hi,
Thanks for sharing a such great work.

My question is regrinding using leewayInSeconds, the default is 5min. Once revoked (i.e.: during password update), a user need to wait 5min before he can contact my server with valid token. Is there a security reason behind this?

Is it ok to remove the following line?(in file: Firebase/Auth.php)
 $tokenAuthenticatedAt = $tokenAuthenticatedAt->modify('-'.$leewayInSeconds.' seconds');

After removing, it works as expected.

jeromegamez:
Fixes #377 

The "validSince" field on a user record marks the boundary before which Firebase ID tokens are considered revoked. This value is compared to the "auth_time" claim of a given ID token.

Until now, we applied the leeway to the "auth_time" value, but not to the "validSince" value, which means that, if both would contain the value `300`, the auth_time would be modified to `0`, but validSince would remain `300`, resulting in the token being considered revoked although it's not.

### How to test

Add/replace in your project's `composer.json` in the `require` section:

```
"kreait/firebase-php": "dev-verification-leeway"
```

:octocat:

### DIFF
--- a/src/Firebase/Auth.php
+++ b/src/Firebase/Auth.php
@@ -550,9 +550,6 @@ class Auth
         $verifiedToken = $verifier->verifyIdToken($idToken);
 
         if ($checkIfRevoked) {
-            $tokenAuthenticatedAt = DT::toUTCDateTimeImmutable($verifiedToken->getClaim('auth_time'));
-            $tokenAuthenticatedAt = $tokenAuthenticatedAt->modify('-'.$leewayInSeconds.' seconds');
-
             if (!($sub = $verifiedToken->getClaim('sub', false))) {
                 throw new InvalidToken($verifiedToken, 'The token has no "sub" claim');
             }
@@ -563,9 +560,17 @@ class Auth
                 throw new InvalidToken($verifiedToken, "Error while getting the token's user: {$e->getMessage()}", $e->getCode(), $e);
             }
 
-            $validSince = $user->tokensValidAfterTime ?? null;
+            // The timestamp, in seconds, which marks a boundary, before which Firebase ID token are considered revoked.
+            if (!($validSince = $user->tokensValidAfterTime ?? null)) {
+                return $verifiedToken;
+            }
 
-            if ($validSince && ($tokenAuthenticatedAt < $validSince)) {
+            $tokenAuthenticatedAt = DT::toUTCDateTimeImmutable($verifiedToken->getClaim('auth_time'));
+            $tokenAuthenticatedAtWithLeeway = $tokenAuthenticatedAt->modify('-'.$leewayInSeconds.' seconds');
+
+            $validSinceWithLeeway = DT::toUTCDateTimeImmutable($validSince)->modify('-'.$leewayInSeconds.' seconds');
+
+            if ($tokenAuthenticatedAtWithLeeway < $validSinceWithLeeway) {
                 throw new RevokedIdToken($verifiedToken);
             }
         }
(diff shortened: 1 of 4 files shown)


tnageleweb:
Hi @jeromegamez,

I tested the modification, and it is worked as expected for me. For this test, I just modified the function (verifyIdToken) at file(Firebase/Auth.php) as seen in the below code snippet. 

**However**, I noticed a major security change, and have a question, in the new changes you validate the clicm('sub') and its user ONLY if $checkIfRevoked is set to true, I'm thinking those are important to be validate regardless whether that token is revoked or not, just an opinion, what do think?

// Modification on file(Firebase/Auth.php)
public function verifyIdToken($idToken, bool $checkIfRevoked = false, /* @deprecated */ bool $allowTimeInconsistencies = null): Token
    {
      
         // @codeCoverageIgnoreStart
        if (\is_bool($allowTimeInconsistencies)) {
            // @see https://github.com/firebase/firebase-admin-dotnet/pull/29
            \trigger_error(
                'The parameter $allowTimeInconsistencies is deprecated and was replaced with a default leeway of 300 seconds.',
                \E_USER_DEPRECATED
            );
        }
        // @codeCoverageIgnoreEnd
        $leewayInSeconds = 300;
        $verifier = $this->idTokenVerifier;
        if ($verifier instanceof IdTokenVerifier) {
            $verifier = $verifier->withLeewayInSeconds($leewayInSeconds);
        }
        $verifiedToken = $verifier->verifyIdToken($idToken);
        if ($checkIfRevoked) {
            if (!($sub = $verifiedToken->getClaim('sub', false))) {
                throw new InvalidToken($verifiedToken, 'The token has no "sub" claim');
            }
            try {
                $user = $this->getUser($sub);
            } catch (Throwable $e) {
                throw new InvalidToken($verifiedToken, "Error while getting the token's user: {$e->getMessage()}", $e->getCode(), $e);
            }
            // The timestamp, in seconds, which marks a boundary, before which Firebase ID token are considered revoked.
            if (!($validSince = $user->tokensValidAfterTime ?? null)) {
                return $verifiedToken;
            }
            $tokenAuthenticatedAt = DT::toUTCDateTimeImmutable($verifiedToken->getClaim('auth_time'));
            $tokenAuthenticatedAtWithLeeway = $tokenAuthenticatedAt->modify('-'.$leewayInSeconds.' seconds');
            $validSinceWithLeeway = DT::toUTCDateTimeImmutable($validSince)->modify('-'.$leewayInSeconds.' seconds');
            if ($tokenAuthenticatedAtWithLeeway < $validSinceWithLeeway) {
                throw new RevokedIdToken($verifiedToken);
            }
        }
        return $verifiedToken;
    }

 Thank you very much.

jeromegamez:
That's a valid point, and I'm glad that you thought about that, and you're right!

I moved the check for the `sub` claim into the verification process, and just pushed the change - thank you for hinting me to that 🌺

The safest option would indeed be to also fetch the user record each time (not only for the revocation check), but this would also cause an additional request to the Firebase APIs each time we verify an ID token. So while you're right, I'm afraid this would/could cause performance issues.

I think: if the token is valid (most importantly: signed by Google) and has a `sub` claim, we can assume with almost certainty that the token is good.

I have plans on adding some more caching layers so that we could also perhaps cache those calls to fetch the user data, but I'm definitely not there yet...

For the time being, with the added check for the `sub` claim and your confirmation that the reported issue is resolved, I hope that we can merge the changes soon and create a new release.

What do you think?

tnageleweb:
Thanks for your great work. Waiting the new release 👍

jeromegamez:
Merged with 6c090e9813cb38ce86e3fe6d42060da22fb2b3c4 and 5f6096a66c2b9242530cb82054259528adb0efc6 🎉

PR #382, opened 2020-02-08
Evaluate unevaluated line breaks in private keys

AliMsayleb:
Sometimes the private key that include a new line inside of it was taking the new line and converting the new line to \n which was causing private key error
This way it will always replace the \n with a new line

### DIFF
--- a/src/Firebase/ServiceAccount.php
+++ b/src/Firebase/ServiceAccount.php
@@ -97,7 +97,7 @@ class ServiceAccount
     public function withPrivateKey(string $value): self
     {
         $serviceAccount = clone $this;
-        $serviceAccount->privateKey = $value;
+        $serviceAccount->privateKey = \str_replace('\n', "\n", $value);
 
         return $serviceAccount;
     }


jeromegamez:
I haven't encountered such a file yet, but it, of course, doesn't mean that it couldn't happen :)

Could you provide me with an example of such a file/value that would cause this problem? I tried re-creating the issue but wasn't able to so far.

AliMsayleb:
> I haven't encountered such a file yet, but it, of course, doesn't mean that it couldn't happen :)
> 
> Could you provide me with an example of such a file/value that would cause this problem? I tried re-creating the issue but wasn't able to so far.

This only happened after i injected the credentials using the array credentials method

jeromegamez:
The thing is: since I can't reproduce the issue (and I tried), an additional test would convince me that it actually exists. If I blindly accept this change and something else breaks, everybody else would be all over me :).

What is the specific error you're getting? Is it an exception or another error? Is it happening while instantiating a `ServiceAccount` instance or later when making a request to the Firebase APIs?

jeromegamez:
I was able to reproduce the issue. It's definitely an edge case, but after much consideration, the change should™ not have negative side effects if the service account credentials are otherwise correctly used.

Thanks for the contribution!

PR #391, opened 2020-03-03
Fixing wrong URL and changed param to string

joao-gsneto:
Fixed Wrong url on this method
change param to string to permit an string containing an customToken to be passed to this method.

This commit fixes this
 method to work with updated to the last google Documentation [see google docs reference](https://firebase.google.com/docs/reference/rest/auth/#section-verify-custom-token)

### DIFF
--- a/src/Firebase/Auth/ApiClient.php
+++ b/src/Firebase/Auth/ApiClient.php
@@ -51,10 +51,10 @@ class ApiClient implements ClientInterface
      * @throws AuthException
      * @throws FirebaseException
      */
-    public function exchangeCustomTokenForIdAndRefreshToken(Token $token): ResponseInterface
+    public function exchangeCustomTokenForIdAndRefreshToken($token): ResponseInterface
     {
-        return $this->requestApi('verifyCustomToken', [
-            'token' => (string) $token,
+        return $this->requestApi('https://identitytoolkit.googleapis.com/v1/accounts:signInWithCustomToken', [
+            'token' => $token instanceof Token ? (string) $token : $token,
             'returnSecureToken' => true,
         ]);
     }


jeromegamez:
Thanks for taking the time to create this PR, I do appreciate it!

If you're willing to rename the commit message to "Update link


The `verifyToken` endpoint is not per se _wrong_, it uses a different base URL (see https://github.com/kreait/firebase-php/blob/master/src/Firebase/Factory.php#L241) and still works, but you're right, why not use the current endpoint.

As for the change of the `Token` parameter, I'd prefer not to have this: this makes it possible to call the method `with `new stdClass()` or `true` and would cause side effects. I know, it's unlikely that someone would do that, especially since this is an internal class (see the `@internal` annotation), only meant to be used from the Auth component, but still 😓

I don't mind the change of URL, but I'd prefer not to change the method argument.

I'd like to ask you to include only the changed URL in the PR and change the commit message to something like `Update custom token exchange endpoint` (without the extended commit message) - it's just that I like the "feeling" of a message like it a bit better.

If you're just interested in the changed URL and don't want to make the changes, please let me know with a short reply, I will then make it myself (and make sure that you still are named as the contributor in the commit history).

Thanks!

jeromegamez:
Merged with https://github.com/kreait/firebase-php/commit/4156d751241895ad2148db04466f6b4bd18af6ae, thanks!

joao-gsneto:
thanks @jeromegamez maibe my english is not good enough. 

So i thing is good to have an method that resolves an idToken from custom token directly, because sometimes this is usefull for an backend API that uses Firebase.
On my case this is necessary to do some integration tests.

PR #398, opened 2020-03-19
Add property salt to UserRecord

bithai:
This exposes the salt property so it's accessible.


jeromegamez:
Merged with dad5c47a057588dfc82653507022f340699c440e, thank you! 🌺

PR #402, opened 2020-04-01
fix links to docs in README file

zulkris:


jeromegamez:
Good catch, thank you for the notification 🙈! I accidentally disabled the `stable` docs on readthedocs.org while preparing the new release, it should be fixed by now!

Issue #404, opened 2020-04-02
Use previously fetched Google Oauth Token


Jeremlicious:
Hi, we use kreait/firebase-php 4.43 in SQS consumer and would like to save some queries to Google.

When doing this :
```php
$firebaseClient->getDatabase()->getReference('path1')->getValue();
$firebaseClient->getDatabase()->getReference('path1')->getValue();
```

the method `Google\Auth\OAuth2::fetchAuthToken` is called twice and two HTTP auth queries are sent to Google.
The received tokens have **expires_in=3599**.

How to use the first token for the second query to Google ?

Jeremlicious:
Hi, thanks ! Yes I will be able to test.
Can't upgrade to 5.0 yet, we are still using php 7.1.19

Jeremlicious:
Hi, thanks !
Here is the tests I made :

```php
$database->getReference('path1')->getValue();
sleep(30);
$database->getReference('path2')->getValue();
```
I get only one call to `Google\Auth\Oauth2::fetchAuthToken`
According to `withEnabledDebug` the same Oauth token is sent ✔️ 

```bash
*   Trying 35.201.97.x...
* Connected to myproject.firebaseio.com (35.201.97.x) port 443 (#0)
* ALPN, offering http/1.1
* Cipher selection: ALL:!EXPORT:!EXPORT40:!EXPORT56:!aNULL:!LOW:!RC4:@STRENGTH
* successfully set certificate verify locations:
*   CAfile: /etc/ssl/certs/ca-certificates.crt
  CApath: /etc/ssl/certs
* SSL connection using TLSv1.2 / ECDHE-RSA-AES128-GCM-SHA256
* ALPN, server accepted to use http/1.1
* Server certificate:
*        subject: C=US; ST=California; L=Mountain View; O=Google LLC; CN=firebaseio.com
*        start date: Feb 20 19:36:06 2020 GMT
*        expire date: Feb 18 19:36:06 2021 GMT
*        subjectAltName: testjerem-92843.firebaseio.com matched
*        issuer: C=US; O=Google Trust Services; CN=GTS CA 1O1
*        SSL certificate verify ok.
> GET /path1.json HTTP/1.1
Host: myproject.firebaseio.com
User-Agent: GuzzleHttp/6.5.1 curl/7.47.0 PHP/7.1.30-1+ubuntu16.04.1+deb.sury.org+1
authorization: Bearer xxxxxxx.SAME.TOKEN.xxxxxxx

[...]

PR #405, opened 2020-04-03
[4.x] Re-use fetched auth tokens in the current process

jeromegamez:
See #404 

:octocat:


jeromegamez:
Merged with fe6f0381dce101a8411b820adb6a1f97346c7faf

PR #406, opened 2020-04-03
[5.x] Re-use fetched auth tokens in the current process

jeromegamez:
See #404 

:octocat:


jeromegamez:
Merged with 7dfcc842f8645fe5238d6936db1f30c8b384a89b

PR #407, opened 2020-04-10
Update cloud-messaging.rst

spike31:
Hi,

In your docs you said 500 messages, but in the code is 100

`SendMessageToTokens`
`SendMessages`


jeromegamez:
Where does it say that? This is the relevant constant:

https://github.com/kreait/firebase-php/blob/master/src/Firebase/Messaging/Http/Request/SendMessageToTokens.php#L18

spike31:
Hmm...Ok my version installed is outdated and I forgot to double check with de master branch!

Sorry :( 

Thx for your work anyway!

PR #409, opened 2020-04-12
Removed Duplicate Import

RahulDey12:


jeromegamez:
This is not a duplicate import 😅  `SendMessage` and `SendMeasages` or two separate classes.

PR #425, opened 2020-05-27
Fix link to Firebase Bundle

rvanlaarhoven:
The original url resulted in a 404 on Packagist, so this should fix that.


jeromegamez:
Whooops 😅!

Thank you! 🌺

PR #440, opened 2020-06-26
New feature: import users

### LINKED ISSUE #388: Is it possible to add support for "Import Users" feature?
I was checking this project and wanted to migrate existing users of a system (with password already hashed and salted) to firebase and looks like firebase has a solution for this-https://firebase.google.com/docs/auth/admin/import-users
But looks like it is not possible with this package.

It would be really nice if we had a method for importing old users in batches just like the examples in firebase documentation.

But my concern is if it is possible? I checked the rest api section of documentation of firebase authentication but found nothing about import user endpoint. If it is not possible can you suggest me any other way to do it in php?
Alternatively I am thinking about creating new user one by one but it seems the `signUpNewUser` endpoint doesn't take any field for hashed password.

TPXP:
Fixes #388 

Thanks for this project!


jeromegamez:
Thank you for taking the time to contribute to the SDK (and even thinking of the documentation 🥰), I highly appreciate it 🌺.

There are a couple of things I would do a bit differently (for example, as you might have seen throughout the SDK I like to wrap inputs in value objects), but I will definitely take your work to build upon and make sure that your contribution is attributed to you as well!

src/Firebase/Auth/ApiClient.php
```diff
@@ -183,6 +184,62 @@ public function unlinkProvider(string $uid, array $providers): ResponseInterface
...
+     * @throws FirebaseException
+     */
+    public function importUsers(array $users, array $options, ProjectId $projectId): ResponseInterface
+    {
+        // Determine the request body
+        $body = [];
+        if (isset($options['hash'])) {
+            if (isset($options['hash']['algorithm'])) {
+                $body['hashAlgorithm'] = $options['hash']['algorithm'];
+            }
```
jeromegamez, line 201: This and the following lines could be simplified with

```php
$body['hashAlgorithm'] = $options['hash']['algorithm'] ?? null;
```

In the end, `null` values could then be filtered out with

```php
$body = array_filter($body, static function ($value) {
    return $value !== null;
});
```

src/Firebase/Auth/ApiClient.php
```diff
@@ -183,6 +184,62 @@ public function unlinkProvider(string $uid, array $providers): ResponseInterface
...
+            ], $userData);
+            unset($res['uid']);
+            // Make sure to pass the hashes in a base64-encoded format
+            if (isset($res['passwordSalt'])) {
+                $res['salt'] = $userData['passwordSalt'];
+                unset($res['passwordSalt']);
+            }
+
+            return $res;
+        }, $users);
```
jeromegamez, line 236: I would change this to:

```php
$body['users'] = [];

foreach ($users as $userData) {
    $userData['localId'] = $userData['uid'] ?? null;
    $userData['salt'] = $userData['passwordSalt'] ?? null;
    unset($userData['uid'], $userData['passwordSalt']);

    $body['users'][] = $userData;
}
```

src/Firebase/Auth/ApiClient.php
TPXP: The static keyword before the callback function puzzles me, I'd love to know more about the difference with

```php
$body = array_filter($body, function ($value) {
    return $value !== null;
});
```

jeromegamez: It's an optimization for lambdas that don't need to reference `$this`. With `static function`, the PHP engine doesn't need to make `$this` available to the function and later garbage collect it.

isteiger:
any update on this?

TPXP:
Please don't close this.

src/Firebase/Auth/ApiClient.php
iampaul: Can we import the users without passwords using this? I have tried to send Facebook provider details like below, but it is not working. How to import existing Facebook member using this function?.

$body['users'] = [[
            localId => 'c9fe1fdb715d05c4c4758e86286',            
            displayName => 'Siva Raj',
            email => 'siva_mpaiefz_raj@tfbnw.net',
            photoURL => 'http://www.example.com/12345678/photo.png',
            emailVerified => true,
            phoneNumber => '+11234567890',
            providerData => [[
              uid => '104653111407187',
              email => 'siva_mpaiefz_raj@tfbnw.net',
              displayName => 'Siva Raj',
              photoURL => 'http://www.example.com/12345678/photo.png',
              providerId => 'facebook.com'
            ]]
        ]];

jseparovic1:
@TPXP @jeromegamez What is the status of PR? Can I help with the implementation to get it done?

IrvinT:
@jeromegamez Any update related to that PR ? I would greatly appreciate to works with

Dirkhuethorst:
@jeromegamez This would also be really helpful when migrating users between tenants. At the moment I do not see another way to do this without losing the credentials

jeromegamez:
This currently hasn't a high priority due to a lack of resources and that I want to implement a few changes and prerequisites before implementing this. For example: in order to be able to test this, I need an "export" feature first - when I can export and re-import the users, I can be confident that everything works.

I do understand that this is important to you all participating/watching this PR and it's a useful one (that's why it's an open PR), but in the meantime, you should still be able to export and import users with the Firebase CLI (https://firebase.google.com/docs/cli/auth).

If this is a crucial feature for your business, please contact me so that I can give you a cost estimate (if we partner up, you'll get a proper invoice in return).

Alternatively (I'm trying something out here) I hereby pledge to start implementing this at the latest on the weekend after I reach my current GitHub Sponsors Goal ($100 per month) 😅

Issue #459, opened 2020-08-10
Remove email address not possible


dododedodonl:
**Describe the bug**
Once an email address is set, it is not possible to remove it.

**To Reproduce using the attributes**
```php
$firebaseAuth->updateUser($uid, ['email'=>null]);
// or
$firebaseAuth->updateUser($uid, ['unverifiedemail'=>null]);
// or
$firebaseAuth->updateUser($uid, ['verifiedemail'=>null]);
```

Results in `TypeError: Argument 1 passed to Kreait/Firebase/Value/Email::__construct() must be of the type string, null given, called in vendor/kreait/firebase-php/src/Firebase/Request/EditUserTrait.php on line 132`

**To Reproduce using the delete attribute**
```php
$firebaseAuth->updateUser($uid, ['deleteattribute'=>'email']);
```

Results in a successful query to the firebase api, without updating / removing anything, the email address is still on the returned `UserRecord`.

**Expected behavior**
`email` attribute to null
`email_verified` attribute to false

**Environment (please complete the following information):**
 - OS: MacOS 10.14.6
 - PHP version: 7.4.8
 - Firebase SDK Version: kreait/firebase-php 5.7.0

**Additional context**
The firebase api will remove the email address if it is supplied in the `deleteAttribute` array on the `UpdateUser` request.

jeromegamez:
Thank you for the detailed issue description 🙌.

This is a property/limitation of Firebase itself and can't be solved in code on the SDK side: as far as I know, it is only possible to change an email address, not to remove it. The only removable attributes are the photo url and the display name.

If you can provide me with documentation or an implementation of working email removal (preferably in the official SDKs, where I couldn't find one), please reopen the issue and/or add another comment pointing me towards it, so I can look into it, thanks!

jeromegamez:
You're fast 😅

jeromegamez:
Implemented with 3af9961fc82774c112138645026a79f8af8e633f

PR #460, opened 2020-08-10
Delete email attribute

dododedodonl:
Implementation to solve #459 

As @jeromegamez mentioned in the issue, this is not in de API documentation. However, the firebase api removes the email address from the user and sets emailVerified to false if this method is used.

When a user has a provider that depends on an email address, the provider is deleted. However, I observed with the password provider, that when a (new) email address is set, the provider becomes active again with the same password. Should the password provider be removed when removing the email address? Or should the email not be removable when a password provider is specified?

### DIFF
--- a/src/Firebase/Request/UpdateUser.php
+++ b/src/Firebase/Request/UpdateUser.php
@@ -13,6 +13,7 @@ final class UpdateUser implements Request
 {
     public const DISPLAY_NAME = 'DISPLAY_NAME';
     public const PHOTO_URL = 'PHOTO_URL';
+    public const EMAIL = 'EMAIL';
 
     use EditUserTrait;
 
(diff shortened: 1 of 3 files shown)


jeromegamez:
Thank you for the PR, this looks good! 🌺

Concerning the re-used password on email re-addition, I'm not sure if this is a question to be answered in the scope of the SDK - the behavior happens in the Firebase backend, we are only™ sending update requests. What do you think?

dododedodonl:
The provider issue is probably out of scope. It could maybe be mentioned somewhere in the docs that something like that can be the case.

jeromegamez:
Merged with 3af9961fc82774c112138645026a79f8af8e633f - thanks again!

PR #482, opened 2020-10-14
feat(auth): Return user email on password reset methods

mladenbl:
In a case where you need to create a custom password reset endpoint and where there is a need for custom logic(related to the local user) that needs to happen after a password reset, currently, there is no way you can connect the local user and Firebase user(without hacking around).

Real-world example:

The user is deactivated(locally) and needs to be activated after the password reset. Because API endpoint(for example), will receive the only oobCode and a new password,  you can't know which user you need to activate without returning email from confirmPasswordReset method or verifyPasswordResetCode(in a case where you want to allow multiple usages of the same code - hopefully not).

### DIFF
--- a/src/Firebase/Auth.php
+++ b/src/Firebase/Auth.php
@@ -615,9 +615,13 @@ class Auth
      * @throws Exception\AuthException
      * @throws Exception\FirebaseException
      */
-    public function verifyPasswordResetCode(string $oobCode): void
+    public function verifyPasswordResetCode(string $oobCode): Email
     {
-        $this->client->verifyPasswordResetCode($oobCode);
+        $response = $this->client->verifyPasswordResetCode($oobCode);
+
+        $email = JSON::decode((string) $response->getBody(), true)['email'] ?? null;
+
+        return new Email($email);
     }
 
     /**
@@ -636,7 +640,7 @@ class Auth
      * @throws Exception\AuthException
      * @throws Exception\FirebaseException
      */
-    public function confirmPasswordReset(string $oobCode, $newPassword, bool $invalidatePreviousSessions = true): void
+    public function confirmPasswordReset(string $oobCode, $newPassword, bool $invalidatePreviousSessions = true): Email
     {
         $newPassword = $newPassword instanceof ClearTextPassword ? $newPassword : new ClearTextPassword($newPassword);
 
@@ -647,6 +651,8 @@ class Auth
         if ($invalidatePreviousSessions && $email) {
             $this->revokeRefreshTokens($this->getUserByEmail($email)->uid);
         }
+
+        return new Email($email);
     }
 
     /**
--- a/src/Firebase/Auth.php
+++ b/src/Firebase/Auth.php
@@ -619,7 +619,7 @@ class Auth
     {
         $response = $this->client->verifyPasswordResetCode($oobCode);
 
-        $email = JSON::decode((string) $response->getBody(), true)['email'] ?? null;
+        $email = JSON::decode((string) $response->getBody(), true)['email'];
 
         return new Email($email);
     }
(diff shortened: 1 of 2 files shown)


jeromegamez:
I absolutely understand why you need this, and while it's a useful change, It would be a backward-compatibility breaking one (thus the failing test) - even if it's unlikely that someone has extended the Auth class to override these methods, it's still possible and so I can't merge this as is :/

I think the best solution would be to introduce two new methods, so it would look something like this:

```php
    /**
     * Verifies the given password reset code.
     *
     * @see https://firebase.google.com/docs/reference/rest/auth#section-verify-password-reset-code
     *
     * @throws ExpiredOobCode
     * @throws InvalidOobCode
     * @throws OperationNotAllowed
     * @throws Exception\AuthException
     * @throws Exception\FirebaseException
     */
    public function verifyPasswordResetCode(string $oobCode): void
    {
        // Not returning the email on purpose to not break BC
        $this->verifyPasswordResetCodeAndReturnEmail($oobCode);
    }

    /**
     * Verifies the given password reset code and returns the associated user's email address.
     *
     * @see https://firebase.google.com/docs/reference/rest/auth#section-verify-password-reset-code
     *
     * @throws ExpiredOobCode
     * @throws InvalidOobCode
     * @throws OperationNotAllowed
     * @throws Exception\AuthException
     * @throws Exception\FirebaseException
     */
    public function verifyPasswordResetCodeAndReturnEmail(string $oobCode): Email
    {
        $response = $this->client->verifyPasswordResetCode($oobCode);

        $email = JSON::decode((string) $response->getBody(), true)['email'];

        return new Email($email);
    }

    /**
     * Applies the password reset requested via the given OOB code.
     *
     * @see https://firebase.google.com/docs/reference/rest/auth#section-confirm-reset-password
     *
     * @param string $oobCode the email action code sent to the user's email for resetting the password
     * @param ClearTextPassword|string $newPassword
     * @param bool $invalidatePreviousSessions Invalidate sessions initialized with the previous credentials
     *
     * @throws ExpiredOobCode
     * @throws InvalidOobCode
     * @throws OperationNotAllowed
     * @throws UserDisabled
     * @throws Exception\AuthException
     * @throws Exception\FirebaseException
     */
    public function confirmPasswordReset(string $oobCode, $newPassword, bool $invalidatePreviousSessions = true): void
    {
        // Not returning the email on purpose to not break BC
        $this->confirmPasswordResetAndReturnEmail($oobCode, $newPassword, $invalidatePreviousSessions);
    }

    /**
     * Applies the password reset requested via the given OOB code and returns the associated user's email address.
     *
     * @see https://firebase.google.com/docs/reference/rest/auth#section-confirm-reset-password
     *
     * @param string $oobCode the email action code sent to the user's email for resetting the password
     * @param ClearTextPassword|string $newPassword
     * @param bool $invalidatePreviousSessions Invalidate sessions initialized with the previous credentials
     *
     * @throws ExpiredOobCode
     * @throws InvalidOobCode
     * @throws OperationNotAllowed
     * @throws UserDisabled
     * @throws Exception\AuthException
     * @throws Exception\FirebaseException
     */
    public function confirmPasswordResetAndReturnEmail(string $oobCode, $newPassword, bool $invalidatePreviousSessions = true): Email
    {
        $newPassword = $newPassword instanceof ClearTextPassword ? $newPassword : new ClearTextPassword($newPassword);

        $response = $this->client->confirmPasswordReset($oobCode, (string) $newPassword);

        $email = JSON::decode((string) $response->getBody(), true)['email'];

        if ($invalidatePreviousSessions) {
            $this->revokeRefreshTokens($this->getUserByEmail($email)->uid);
        }

        return new Email($email);
    }
```

Would this be a solution that you could work with? If you have better method names, please let me know 😅.

jeromegamez:
Please have a look at #485 and let me know what you think 🙏

mladenbl:
@jeromegamez I agree with you, and I understand your concerns, like you said, even unlikely no-one extended Auth class, it's still possible, so I'm fine with the proposed solution. Thanks!

PR #488, opened 2020-10-22
Make message data semi-stringable

trbsi:
Can we make so our Message Data accepts integers not only strings?

### DIFF
--- a/src/Firebase/Messaging/MessageData.php
+++ b/src/Firebase/Messaging/MessageData.php
@@ -23,8 +23,8 @@ final class MessageData implements \JsonSerializable
         $messageData = new self();
 
         foreach ($data as $key => $value) {
-            if (!self::isStringable($key) || !self::isStringable($value)) {
-                throw new InvalidArgumentException('Message data must be a one-dimensional array of string(able) keys and values.');
+            if (!self::isStringableAndNumeric($key) || !self::isStringableAndNumeric($value)) {
+                throw new InvalidArgumentException('Message data must be a one-dimensional array of string(able) or numeric keys and values.');
             }
 
             if (self::isBinary((string) $value)) {
@@ -34,7 +34,7 @@ final class MessageData implements \JsonSerializable
                 );
             }
 
-            $messageData->data[(string) $key] = (string) $value;
+            $messageData->data[$key] = $value;
         }
 
         return $messageData;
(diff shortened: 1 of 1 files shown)


jeromegamez:
Thank you for taking the time to contribute, it's highly appreciated! 🙌 

Unfortunately, this change would not change the messages that are received in the client - Firebase expects the data field names and values to be strings (see https://firebase.google.com/docs/reference/fcm/rest/v1/projects.messages#Message.FIELDS.data) - and if they are not strings, Firebase will convert them to strings before delivering them to the clients.

```php
$messaging->send(CloudMessage::fromArray([
	'token' => $token,
	'notification' => [
		'title' => 'Title',
		'body' => 'Body',
	],
	'data' => [
		1 => 1
	]
]));
```

Result:
image

src/Firebase/Messaging/MessageData.php
```diff
@@ -51,9 +51,9 @@ public function jsonSerialize(): array
     /**
      * @param mixed $value
      */
-    private static function isStringable($value): bool
+    private static function isStringableAndNumeric($value): bool
     {
-        return \is_null($value) || \is_scalar($value) || (\is_object($value) && \method_exists($value, '__toString'));
+        return \is_numeric($value) || \is_null($value) || \is_scalar($value) || (\is_object($value) && \method_exists($value, '__toString'));
```
jeromegamez, line 56: `is_numeric()` is already covered by `is_scalar()`

src/Firebase/Messaging/MessageData.php
```diff
@@ -51,9 +51,9 @@ public function jsonSerialize(): array
     /**
      * @param mixed $value
      */
-    private static function isStringable($value): bool
+    private static function isStringableAndNumeric($value): bool
```
jeromegamez, line 54: The method would have to check if the value is numeric to make sense: with a value of `"string"`, the method would still return `true`

src/Firebase/Messaging/MessageData.php
```diff
@@ -34,7 +34,7 @@ public static function fromArray(array $data): self
                 );
             }
 
-            $messageData->data[(string) $key] = (string) $value;
+            $messageData->data[$key] = $value;
```
jeromegamez, line 37: As explained in https://github.com/kreait/firebase-php/pull/488#issuecomment-714423182, the Firebase API does want field names and values to be strings, and will convert non-strings to string when delivering the message to the clients.

trbsi: I see, I didn't know that. Thanks for the review

trbsi: For that reason I'll close PR

jeromegamez: Thank you nonetheless! 🌺

Issue #394, opened 2020-03-11
Add a tenant-aware request type for getting user credentials


Pickeringwareltd:
**Is your feature request related to a problem? Please describe.**
I have a multi-tenanted authentication setup, I realise that this is handled within GIP however, there is very little support for it with PHP and the rest of my application uses firebase, as I'm sure other developers will too. 

**Describe the solution you'd like**
I'd like to add the tenantId to the appropriate requests, so that the tenant is taken into consideration when accessing user information, preventing a 'User not found error'. This is very easy to achieve: 

An example when getting user credentials via UID:
Add a nullable tenantID parameter to Firebase\Auth.php@getUser 
Add a nullable tenantID parameter to Firebase\Auth\ApiClient@getAccountInfo
Add the tenantId into the JSON request body like so:
```
return $this->requestApi('getAccountInfo', [
            'tenantId' => $tenantId,
            'localId' => $uid
        ]);
```

A null tenant ID defaults to the standard set of users, meaning this then allows tenant-aware clients to authenticate their users, whilst not affecting the current users who use it without multi-tenanted applications. 

**Describe alternatives you've considered**
I've looked around for a PHP SDK written for the  GIP but there is very little around, this seems to be the most concise platform for authentication.

jeromegamez:
Thank you for this feature request, I‘ll look into it and hope that I will come up with a good implementation.

Please note that I will soon be dropping support for PHP <7.4 (probably after #390), so this feature will not be provided as a 4.x release.

Release 5.x will be fully compatible to 4.x though (except the PHP requirement and the removal of deprecated methods and classes), so if you consider using this SDK and are on a current PHP version, everything should be fine 😅

martijnvdgrift:
Hey @jeromegamez ,

We're running into the same issue. We're storing our users in specific [Tenants](https://cloud.google.com/identity-platform/docs/multi-tenancy), but we're unable to delete this because this library isn't "tenant-aware". Trying to remove a user results in a `User not found` error.

Is this feature request still on your radar?

The "official" NodeJS library does have support for this, so maybe it can be used as inspiration.

Thanks!

jeromegamez:
Hey @Pickeringwareltd and @martijnvdgrift - better late than never: Tenant awareness has landed with [release 5.12.0](https://github.com/kreait/firebase-php/releases/tag/5.12.0) 🥳🚀

PR #493, opened 2020-10-29
Make ApiClient tenant aware 

mostafakram:
Working on fix for https://github.com/kreait/firebase-php/issues/394

I am aware we should have discussed this before adding the pull request but my current development requires my project to be tenant aware and i needed a solution so what do you think ? 

My inspiration was from the [docs](https://cloud.google.com/identity-platform/docs/multi-tenancy-authentication) itself where from the JS sdk they set tnenat like this 
```javascript
firebase.auth().tenantId = 'TENANT_ID';
```

so i thought we can hava similar approach by calling 

```php

$auth.setTenantId($tenantId)
```

or even quicker with laravel `app()` 

```php
app('firebase.auth')->setTenantId($tenantId);
```

and this method will add the required 'tenantId' parameter to the body sent to requests in `requestApi` of `ApiClient`

### DIFF
--- a/src/Firebase/Auth.php
+++ b/src/Firebase/Auth.php
@@ -907,7 +907,7 @@ class Auth
         return $this->getUser($uid);
     }
 
-    public function setTenantId($tenantId)
+    public function setTenantId(string $tenantId): Auth
     {
         $this->client->setTenantId($tenantId);
 
--- a/src/Firebase/Auth.php
+++ b/src/Firebase/Auth.php
@@ -906,4 +906,11 @@ class Auth
 
         return $this->getUser($uid);
     }
+
+    public function setTenantId($tenantId)
+    {
+        $this->client->setTenantId($tenantId);
+
+        return $this;
+    }
 }
--- a/src/Firebase/Auth/ApiClient.php
+++ b/src/Firebase/Auth/ApiClient.php
@@ -29,6 +29,9 @@ class ApiClient implements ClientInterface
     /** @var AuthApiExceptionConverter */
     private $errorHandler;
 
+    /** @var String */
+    private $tenantId;
+
     /**
      * @internal
      */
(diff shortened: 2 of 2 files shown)


mostafakram:
@jeromegamez what do you think ? 

can we have such an implementation 

or can we make the client public to the api so that we can override it ?

jeromegamez:
Unfortunately, this change breaks the principle of having everything configured through the factory and of abstracting away the underlying HTTP communication. Injecting the tenant into the API client also addresses just a part of the tenant awareness: custom token generation, ID token verification and tenant management still would need to be handled separately. Merging this would make it harder vor me to maintain the SDK going forward, so I'm sorry to tell you I can't merge this.

The JS SDK being a client SDK does things differently, and it's principles don't necessarily apply to the Admin SDKs. The Admin Node SDK doesn't do this either, but has a TenantManager and a TenantAwareAuth. I'm not sure yet I can take a similar approach with the PHP Admin SDK, but this SDK doesn't aim for 1:1 similarity anyway, and it has more legacy and doesn't have as many major releases as the official Admin SDKs as well.

I'm currently working on introducing tenant awareness in a different way, and can publish the work as a draft PR/branch in the coming days, but I can't give an estimate when it will make it into a release.

If you're in dire need of this, you could instantiate the Auth component without the factory (or override the factory) and pass an extended ApiClient with your changes - as it is internal, you should pin the kreait/firebase-php to the current version, so that a next release doesn't break your override. Once I released the tenant awareness, you could then migrate back.

I'm sorry that this is not the answer you were looking for, but my focus with the SDK is on maintainability and stability - I do realize that this means that new features are realized much slower than the official SDKs, but other than Google, I don't generate any income with this project, thus my priorities are different 😅

jeromegamez:
I'll add a comment to let you know when the draft PR is ready.

mostafakram:
@jeromegamez thank you for your response 

Nothing to be sorry about at all. I totally agree with you and as i was just proposing and idea for the solution not the solution :D 

As for me i think will workaround firebase for multitenant and i dont think i will have issues now with the current implementation 

Looking forward for your PR regarding this 

Thanks again and good luck ^_^

jeromegamez:
Thank you for your understanding 🌺 and good wishes 😅 ! I've started with the Draft PR at https://github.com/kreait/firebase-php/pull/496, in case you want to keep an eye on it ^^

PR #503, opened 2020-11-22
InvalidToken catch block is never reached

christiaan:
InvalidToken extends InvalidArgumentException so the catch block will never be reached


jeromegamez:
Thanks Christiaan! 🌺

PR #536, opened 2020-12-31
Bug fix: subscribe promise rejected state processing

itkreydo:


jeromegamez:
Good catch, thanks! 🌺 

Merged with f468d13fe5950add4a065fc2a11e9a03dba461de

PR #575, opened 2021-03-08
Add screenName to providerData to get Twitter handle

kingflamez:
Getting Twitter handle is not currently possible with the library, added the handle to the providerData to make it accessible


jeromegamez:
Could you show me a documentation where the `screenName` is referenced? I was just looking at the NodeJS Admin SDK and can't find a field `screenName`, only `displayName`.

jeromegamez:
Merged with 74f663cd29b03a79b89ee426cbd57297510cea1c - thank you for your contribution, and sorry it took me so long (I created a Twitter developer account to test this just today 😅)

PR #578, opened 2021-03-15
Make giggsey/libphonenumber-for-php an optional dependency

jeromegamez:
See #577


acasademont:
@jeromegamez looks good to me, that was fast! I wasn't sure if you preferred to throw an exception if the library was not installed, but given the size of the library, I think it makes sense to make it 100% optional.

jeromegamez:
An exception would have caused a breaking change and I'm not ready for a new major release yet 😅

Some might argue that this is also a breaking change, when a project uses the library as a transient dependency, but I already have a response prepared for that 😬

I'll merge the PR and create a new minor release soon 💪

acasademont:
Awesome, thanks :D

PR #625, opened 2021-06-16
Support for Database Auth Variable Override

tomykho:


src/Firebase/Database/ApiClient.php
```diff
@@ -20,14 +20,16 @@ class ApiClient
 {
     private ClientInterface $client;
     protected DatabaseApiExceptionConverter $errorHandler;
+    private ?string $authVariableOverride;
 
     /**
      * @internal
      */
-    public function __construct(ClientInterface $httpClient)
+    public function __construct(ClientInterface $httpClient, ?array $authVariableOverride)
```
jeromegamez:
```suggestion
    public function __construct(ClientInterface $httpClient, ?array $authVariableOverride = null)
```

src/Firebase/Factory.php
```diff
@@ -168,10 +170,11 @@ public function withDisabledAutoDiscovery(): self
     /**
      * @param UriInterface|string $uri
      */
-    public function withDatabaseUri($uri): self
+    public function withDatabaseUri($uri, ?array $authVariableOverride): self
```
jeromegamez:
```suggestion
    public function withDatabaseUri($uri, ?array $authVariableOverride = null): self
```

src/Firebase/Database/ApiClient.php
```diff
@@ -185,6 +187,12 @@ private function requestApi(string $method, $uri, ?array $options = null): Respo
 
         $request = new Request($method, $uri);
 
+        if ($this->authVariableOverride) {
+            $options['query'] = [
+                'auth_variable_override' => $this->authVariableOverride,
+            ];
+        }
```
jeromegamez: I'm not sure how Guzzle query parameter options work, but it could be that the `$uri` already has query parameters present (e.g. when preforming a realtime database query). 

Here, the existing query parameters should be extracted from the incoming `$uri`, the `auth_variable_override` should be added, and then replace the query on the `$uri` with the now extended set of query parameters.

Instead of setting them with Guzzle options because there could already be other query parameters present (e.g. because the Request comes from a Realtime Database Query)

jeromegamez (changes requested):
Thank you for taking the time and putting work into the PR, it's much appreciated 🌺 !

I added some notes to some things that stood out to me at first sight, I will have a deeper look as soon as I can.

src/Firebase/Database/ApiClient.php
tomykho: Fixed in https://github.com/kreait/firebase-php/pull/625/commits/39881566da97c54135ad74e26ba5176cfa4a754c.

jeromegamez:
Merged with a7fbc1a38791a94293a4ff42c2a52b0b0e1c3d9b, tested and documented with 16a654c856c44c4869c1eba9bb9d3337a6ed5717

In the end, I took a different approach than you proposed, but made sure that you remained as the author of the feature.

The reason why I went with a middleware instead of the additional parameter to the `withDatabaseUri()` is that there are three different modes that can apply:

* the database is accessed with admin privileges (default)
* the database is accessed with _limited_ privileges (what you proposed)
* the database is accessed with **no** privileges at all (what's needed as well)

For this reason, just having an additional `authVariableOverride` property in the factory isn't enough, because it being `null` would be ambiguous - we can't determine if that's meant to be "use the default administrative privileges" or "use no privileges at all".

Long story short, the feature is in and documented here: https://firebase-php.readthedocs.io/en/latest/realtime-database.html#authenticate-with-limited-privileges

Thanks again for your contribution, and also for your patience! 🌺

PR #630, opened 2021-06-30
Add Identity Platform Endpoints

Ikstar:
Google has added a new layer for Firebase Auth under Identity Platform, we needed a number of the Identity Provider endpoints to work for our application. Let me know, if this is something that would be welcome here as an addition.

https://cloud.google.com/identity-platform/docs/reference/rest


jeromegamez:
Woah, that's huge! In terms of work you put in and also in terms of the amount of code to look through 😅. First and foremost: thank you for proposing this PR!

You'll hopefully understand that this isn't something that I can merge right away, and that I will probably at least have some change requests that _might_ not be to your liking 😅 - just wanted to make clear in advance that it's not personal.

I say this before having a look at it in detail and testing it out - my time is quite limited at the moment so I can't give you a timeline on when I can tackle this, so I hope you already have it running in your application so you don't require it to be in the SDK itself.

That being said, I noticed a couple of things while skimming over the files:

* I know it removes comfort, but please don't include `nesbot/carbon` as an additional dependency. The SDK has a `DT`helper to convert values to `DateTimeImmutable` objects (https://github.com/kreait/firebase-php/blob/5.x/src/Firebase/Util/DT.php) 
* Could you please run PHP CS Fixer on the code - it's usually not a requirement for PRs, but it would make the review a little easier in this case (especially for PHPDocs).

Thank you!

Ikstar:
Thank you for considering the addition. There is no hurry in regards to having this merged in on our end, time is limited for everyone. 

I tried to follow similar structure to Auth section of the project. I am sure there are a number of changes that would make sense to be made, I am not sold on all the decisions in there.
- I mimicked the objects that Google specified in their spec. (With hope once they get the GRPC wrapper for these APIs in, it would be dead simple to have a GRPC handler instead of GuzzleHandler)
- Not sure if Request and Response Objects need to be separate classes

* Removing Carbon is not an issue - Done
* php-cs-fixer - Done (was wondering why phpdocs didn't get formatted)
     * (Added .gitkeep for the tools/php-cs-fixer directory, php-cs-fixer fails without the cache directory available)

We use Laravel, so I do have a companion PR for laravel-firebase for down the road.

jeromegamez:
Oh, that's the one thing that I wanted to mention but forget in my previous comment - I've seen that you mimicked the other components of the SDK and appreciate it! (Although some of the other components have become quite messy over time 😂) Thank you!

I'm going to release the workflows so that you can evaluate their results 💪

Ikstar:
Clearly, I missed support for php 7.4, I'll get those issues addressed as well as adding unit testing to objects.

Pretty sure Github Actions restricts access to Secrets to forked PRs, so the integration tests didn't get credentials.

jeromegamez:
No problem, once we're ready, I will run the integration tests locally, so let's ignore them together with the code coverage 💪

src/Firebase/IdentityPlatform.php
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function listAllDefaultSupportedIdpConfigs(): array
+    {
+        $response = $this->client->listAllDefaultSupportedIdpConfigs();
+
+        return ArrayHelper::flatten($this->getResponseAsArray($response));
```
jeromegamez: Would it be possible to integrate the flattening into the class and remove the `ArrayHelper`? I believe I've seen it used only here… even if not, I don't mind code duplication here and there. (I know, that's a bold statement considering I introduced the `DT` helper back in the time, but please bear with me 😅)

src/Firebase/IdentityPlatform.php
```diff
@@ -0,0 +1,190 @@
...
+    {
+        foreach ($x as $arg) {
+            if ($arg instanceof ApiClient) {
+                $this->client = $arg;
+            }
+        }
+    }
+
+    /**
+     * {@inheritDoc}
```
jeromegamez: `@inheritDoc`s are not needed when they're the only thing in the PHPDoc 🙏

src/Firebase/IdentityPlatform.php
```diff
@@ -0,0 +1,190 @@
...
+class IdentityPlatform implements Contract\IdentityPlatform
+{
+    private ApiClient $client;
+
+    /**
+     * @param iterable<ApiClient|null>|ApiClient|null ...$x
+     *
+     * @internal
+     */
+    public function __construct(...$x)
```
jeromegamez: I know I did this in the `Auth` class, but this was only because it was one of the oldest classes and I needed it to be backward compatible at the time - you can inject the dependencies directly here without using the `foreach` construct 🙏

Ikstar:
- Added some Unit tests for the objects. 
- Removed Array Helper
- Cleaned up some of the docblocks
- Fixed 7.4 compatibility
- Addressed existing comments

jeromegamez:
I'm sorry for being slow to react, I currently have a lot going on. I checked out the code locally and called some methods and looked at the code and there are some things that caught my eye that I would like to have different in order to feel comfortable with them - but we should get to that once the Github actions run through without errors (the code coverage doesn't count, I use it only as an indicator/for its informational value, it's not a hard requirement).

(And please don't forget that you can stop working on this at any time - I value the work you've already put into it, and if you decide to let it rest, I will still make sure that you are attributed to the feature as a co-author)

Ikstar:
I haven't had too much time to work on this myself last couple of weeks. 

- Got the Github actions working in my fork, so the unit tests and phpstan pass now.

jeromegamez:
I managed to have a deeper look and I must admit that I don't understand half it, I'll need to dig deeper, sorry again for the long wait! There are some things I noticed but I can change them on my own (it's just programming style/naming stuff so not really relevant).

This brings me to two approaches that we could take from here, and I'm looking forward to knowing which you'd prefer:

One approach would be that I comment all the things that stood out to me or that I would change to feel more comfortable with maintaining down the road (including nitpicky details 🙈) and ask questions where I don't understand what's happening. The drawback here is that I don't want to cause you even more work or negative feelings because of the nitpickyness.

The second one would be that I take over in a new branch/PR directly here in the repo and apply "my" changes as I go and learn on my own how everything works. This has the drawback that you would have to follow that new branch to be aware of those changes (names, method signatures, etc.).

What do you think?

MGatner:
If @Ikstar has "Allow Edits from Maintainers" checked on the PR then you can add his fork as a remote and collaborate on the PR directly from his branch. I do this a lot at our framework repo to touch up PRs.

Ikstar:
Unfortunately because the repository is in an org the option to Allow Edits from Maintainers is not available. I invited you as a writer on the repository.

Probably a combination of both first and second would work fine. The project that uses the functionality is actively under development for us, so I can make time to partake in this endeavor. 

Gladly, can go over and explain any parts that don't make sense.

PR #635, opened 2021-07-07
Added support for linking IdP to existing account

stanogurnik:
Added support for linking IdP Provider as described in REST API ref. https://firebase.google.com/docs/reference/rest/auth/#section-link-with-oauth-credential 
No tests added as reliable integration testing is nearly impossible.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 ### Added
 * Added support for Session Cookie Generation
   ([Documentation](https://firebase-php.readthedocs.io/en/latest/authentication.html#session-cookies))
+* Added support for linking Idp to existing account
 
 ### Changes
 * Bumped `kreait/firebase-tokens` to `^1.16`
--- a/docs/authentication.rst
+++ b/docs/authentication.rst
@@ -287,6 +287,19 @@ Sign In without a token
     $signInResult = $auth->signInAsUser($userOrUid, array $claims = null);
 
 
+Linking and Unlinking IdP
+-------------------------
+
+.. code-block:: php
+
+    $auth->unlinkProvider($uid, $provider)
+
+For linking IdP you can use any of above methods for signing in with IdP credentials, by providing linking ID token
+
+.. code-block:: php
+
+    $signInResult = $auth->signInWithFacebookAccessToken($accessToken, null, $idToken);
+
 ************************
 Invalidate user sessions
 ************************
--- a/src/Firebase/Auth.php
+++ b/src/Firebase/Auth.php
@@ -569,22 +569,22 @@ class Auth implements Contract\Auth
         throw new FailedToSignIn('Failed to sign in anonymously: No ID token or UID available');
     }
 
-    public function signInWithTwitterOauthCredential(string $accessToken, string $oauthTokenSecret, ?string $redirectUrl = null): SignInResult
+    public function signInWithTwitterOauthCredential(string $accessToken, string $oauthTokenSecret, ?string $redirectUrl = null, ?string $linkingIdToken = null): SignInResult
     {
-        return $this->signInWithIdpAccessToken(Provider::TWITTER, $accessToken, $redirectUrl, $oauthTokenSecret);
+        return $this->signInWithIdpAccessToken(Provider::TWITTER, $accessToken, $redirectUrl, $oauthTokenSecret, $linkingIdToken);
     }
 
-    public function signInWithGoogleIdToken(string $idToken, ?string $redirectUrl = null): SignInResult
+    public function signInWithGoogleIdToken(string $idToken, ?string $redirectUrl = null, ?string $linkingIdToken = null): SignInResult
     {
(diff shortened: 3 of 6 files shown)


jeromegamez:
Hey, and thank you for taking the time to implement this! This functionality was already present in the 4.x releases of the software and didn't make the cut for 5.x because it seemed that nobody was using it, so I'm happy you're bringing it back!

This PR looks good, and I'm glad that you added docs and noted why no integration tests are included (and you're right!). However, I do feel that linking an IdP to an existing account should be a separate action/method, even if the `signIn*` methods all use the same REST endpoint 🤔.

That would of course mean that two steps would be needed: first signing in the user with the provided IdP credential, and when the sign-in succeeded, linking the accounts.

```php
$idTokenOfTargetUser = '...'; // e.g. from a previous sign in with email/password

$idpSignIn = $auth->signInWithIdpAccessToken($provider, $accessToken);

$linkingResult = $auth->linkAccounts($provider, $idpSignIn->idToken(), $idTokenOfTargetUser);
```

What do you think? If you can convince me that adding yet another parameter to the `signIn*` methods is justified, I'll gladly accept 😅, but at the moment I really think these two operations should be split.

(If you say: "okay, but I don't want to spend even more time on this than I already have", I will do it based on what you already provided and make sure that you're immortalized as the co-auther of the commit(s) related to this feature)

stanogurnik:
Hey, thank you for response, I was thinking the same as you, that linking/unlinking provider should be separate method (as unlinking currently is), altho signature of rest endpoint suggest that it relates somehow to signing in. I mean I was not happy with implementation but given current design it seems like lesser evil. Main reason why to reuse signin method was that I not wanted to reimplement whole signature to enable this, but as you pointed out, using ID token from signed provider instead of access token from provider itself simplifies signature a lot. i will refactor this PR, thank you for your comments.

stanogurnik:
Hi, I tried to implement your solution which is doable but unfortunately it does not solve our scenario: We have multiple sign in providers all linked to one account, which is fine, but we don't want to have duplicate identities in firebase so we have "One account per email address " (which is default by reason). So in out case, user has already account with password provider and want to add facebook/apple login (that have same email). So you already can't sign in (and get id token) user with idp access token. When yo try that in firebase, it will return you user data provided by facebook, but no ID Token. In this scenario only doable solution is to link idp using access token. Unfortunately current design does not allow other solution only adding new parameter to method's signature. 

In future major release i would think of letting developer build SignIn action object and passing that instead of parameters. that way you can extend transparently method's signature without adding more parameters. 

```php
$action = SignInWithIdpCredentials::withIdToken(new Provider(Provider::GOOGLE), $idToken)
    ->withRequestUri($redirectUrl);
    // ... potentially more attributes in future
    
$signInResult = $auth->signInWithAction($action);
```

jeromegamez:
Thank you for the detailed explanation and the insights into the use case (I rarely get these and don't know how and where the SDK is used in general 😅). I'll release the workflows now and have a closer look hopefully soon! Thanks again! 🙏

jeromegamez:
I'm curious and it's totally possible that I'm slow to understand - even when the target ID token is used as the last parameter in the `signInWith*` method, how do you retrieve the target ID token in the first place? So, assuming you already have an existing email/password user in your auth database. When you link a Facebook user to an existing Firebase user, how do you know that the "new" credentials belong to the existing user?

So, I think what I mean is - if you do a `$auth->signInWithFacebookAccessToken($accessToken, null, $idToken)` - how do you know the `$idToken` in the first place?

(This is all really for me to understand the account linking process)

stanogurnik:
@jeromegamez yes for linking IdP user has to be logged in first so you already have IdToken. There are multiple different scenarios that need to covered, but generally, if user want to link accounts, he need to be logged in first at least to one of them. 

To have better understanding on all the combinations and scenarios in user flow: 
- user already have account (password) and want to add facebook login (with same email). When he try to use that without login in he is returned back with error message that he needs to login in before using IdP to login. (then he logs in he is able to connect provider even with same email)
- user already have account (password) and want to add facebook login (with different email). When he try to login, user is redirected to account creation page. If his goal was to link account, he leaves sign up flow, log in and link provider to his account

There are more, but those should get you the idea :) hope this helps

jeromegamez:
I'm sorry for being slow to react, I currently have a lot going on. I rebased your PR with the latest updates from the 5.x branch and the [tests are successfully executed](https://github.com/kreait/firebase-php/actions?query=branch%3Alink-user-to-idp-credentials++).

Could you check the branch out and confirm that the operations work as expected? Unfortunately, I don't have the means to test this "for real" myself…

stanogurnik:
@jeromegamez no problem at all. Rebased fork agains 5.x branch and resolved conflict. Tested fork manually and works as expected.

jeromegamez:
Merged with slight modifications with fc9933f7b6fa4ee011384fc902f84eba944ce1d6

Thanks again @stanogurnik (also for your patience) 🌺

PR #645, opened 2021-08-07
Fix array shapes of AndroidConfig

hoogi91:
The TTL has to be defined as "string" (Duration) instead of "int|double" as mentioned in referenced documentation https://firebase.google.com/docs/reference/fcm/rest/v1/projects.messages#androidconfig

### DIFF
--- a/src/Firebase/Messaging/AndroidConfig.php
+++ b/src/Firebase/Messaging/AndroidConfig.php
@@ -18,7 +18,7 @@ final class AndroidConfig implements JsonSerializable
     /** @var array{
      *      collapse_key?: string,
      *      priority?: 'normal'|'high',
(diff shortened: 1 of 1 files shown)


jeromegamez:
Thanks Thorsten! The field is also referenced in the `fromArray()` method. FYI: I would have liked using PHPStan's type aliases, but PHPStorm doesn't support them (yet), that's why the shapes are duplicated so often 😅

hoogi91:
Hey @jeromegamez :) yep missed that one 😄

jeromegamez:
Merged with https://github.com/kreait/firebase-php/commit/76c6dd381e44206cca97bec9f8abf39856365b5d, thanks again! 🌺

PR #650, opened 2021-08-23
Add function to import users & batch delete users

sondnpt00343:
Thanks for a great library 👏👏👏

### DIFF
--- a/docs/user-management.rst
+++ b/docs/user-management.rst
@@ -292,6 +292,25 @@ Delete a user
 
     $auth->deleteUser($uid);
 
+*************
+Delete many users
+*************
+
+.. code-block:: php
+
+    $body = [
+        // Required. List of user IDs to be deleted.
+        'localIds' => [
+            'some-uid-1',
+            'some-uid-2',
+            'some-uid-3',
+        ],
+        // Whether to force deleting accounts that are not in disabled state. If false, only disabled accounts will be deleted, and accounts that are not disabled will be added to the errors.
+        'force' => true,
+    ];
+
+    $auth->deleteUsers($body);
+
 ************************
 Using Email Action Codes
 ************************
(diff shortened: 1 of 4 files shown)


jeromegamez:
Thank you for taking the time to contribute to the SDK, I very much appreciate it, especially that you even thought about adding documentation 😍.

While your implementation works perfectly fine (the code looks good and I tested the `deleteUsers()` method on my test projects), I do belive that we can do it a little more convenient for the users of the SDK 😅 

As an example, you propose this method call to batch delete users:

```php
$options = [
    'localIds' => [
        'some-uid-1',
        'some-uid-2',
        'some-uid-3',
    ],
    'force' => true,
];

$auth->deleteUsers($options);
```

This is identical to the payload that the Firebase REST API expects to receive, but it opens the room for problems - for example, the developer could misspell `localIds` and write `local_ids` instead. On the other side, if Google decided to change the expected payload, everybody using the SDK would need to update their code to match the new payload requirements.

I would like the following method signature better:

```php
$uids = ['some-uid-1', 'some-uid-2', 'some-uid-3';
$forceDeleteEnabledUsers = true;

$auth->deleteUsers($uids, $forceDeleteEnabledUsers);
```

Based on this, I implemented the batch deletion of users as you can see in https://github.com/kreait/firebase-php/commit/76c2f40d36c26552c69c546cc12031f5d68b591c and added some elements to make the developer experience less "surprising" 😅:

* The project ID is optional, but the `deleteUsers()` cannot be used when it's not present. In that case, an exception is thrown with a little bit of detail why it has been thrown
* The response of a batch deletion is parsed into a `DeletedUsersResult` object which can be used to retrieve the success count, failure count and errors, if available.
* The `DeleteUsersRequest` checks if the UIDs are valid and if they're not too many (batch requests are limited to 1000 UIDs)
* Some unit and integration tests

Since you laid the groundwork for this new feature, I dedicated the commit to you and added you as the co-author, I hope that's okay 🙌.

Thanks again for introducing this and motivating me to bring it into the main branch!

jeromegamez:
As for the batch import of users... I would probably give this feature a similar treatment, although it's far more complicated, technically (I need to feel comfortable with the implementation in order to continue to maintain it) as well as "on principle" 😅 

You've certainly seen that the batch import has been proposed/requested last year with a similar approach (https://github.com/kreait/firebase-php/pull/440), but I didn't feel comfortable with the implementation back then either.

It seems to be a highly anticipated feature, but nobody took me up on my offer to make it happen (https://github.com/kreait/firebase-php/pull/440#issuecomment-797396703):

> If this is a crucial feature for your business, please contact me so that I can give you a cost estimate (if we partner up, you'll get a proper invoice in return).
> 
> Alternatively (I'm trying something out here) I hereby pledge to start implementing this at the latest on the weekend after I reach my current GitHub Sponsors Goal ($100 per month) 😅
> 
> https://github.com/sponsors/jeromegamez

Now, since the offer still stands, I "can't™" implement this feature on my own until I have reached the goal, I hope you understand 😅 - however, if you want to give it a go in a new PR, I'll add my code reviews. I can already tell you that my target implementation would look something like this:

```php
public function importUsers(UserImportRecords|array $userImportRecords, UserImportOptions|array $importOptions)
```

with arrays being parsed to the respective classes, user record validations, request/result objects, integration tests, documentation 🙈. It's a big feature ^^

Issue #654, opened 2021-09-08
Support nonce for Apple login


chyt:
### Describe the feature you would like to see

When performing login with Apple, there are cases where a [nonce](https://developer.apple.com/documentation/sign_in_with_apple/clientconfigi/3386668-nonce) is required in the response body to associate a client session and an ID token. I don't believe this option is yet supported by this library. 

An implementation of this can be in my fork [here](https://github.com/watchcharts/firebase-php/commit/9869fcf79acf84d51b215c1d0fa1009306565c22?w=1). However it was written a while back and not compatible with the current codebase.

jeromegamez:
Thanks @chyt! This seems pretty straight forward thanks to your fork, I'll try to get it to the main branch shortly 🤞

jeromegamez:
Hey @chyt, could you give https://github.com/kreait/firebase-php/pull/658 a try and confirm that it works as expected? Unfortunately I can't really test this myself and can only work on the assumption that I implemented the requirements correctly 🤞

chyt:
@jeromegamez I will give this a try in the next few days and report back.

PR #658, opened 2021-09-11
Add support for providing a nonce when signing in with IdP credentials

jeromegamez:
Implements #654 

:octocat:


MGatner:
TIL "nonce" 🧐

PR #659, opened 2021-09-19
(WIP) Add support for using the Auth simulator

jeromegamez:
⚠️ Dirty branch, much junk code, parts might be broken, force pushes, you name it 😅 

### How to use

* Install the Firebase CLI Tools (`npm install -g firebase-tools`)
* Set the `FIREBASE_AUTH_EMULATOR_HOST` env variable to `localhost:9099`
* In your project dir, execute `firebase init emulators`
* Start the emulator suite with `firebase emulators:start`

### Example

```php
<?php

declare(strict_types=1);

require 'vendor/autoload.php';

use Kreait\Firebase;

\putenv('FIREBASE_CREDENTIALS=/path/to/firebase_service_account.json');
\putenv('FIREBASE_AUTH_EMULATOR_HOST=localhost:9099');

$factory = new Firebase\Factory();

$auth = $factory->createAuth();

try {
    $user = $auth->createAnonymousUser();
} catch (Throwable $e) {
    echo $e->getMessage();
    exit;
}

echo json_encode($user, JSON_PRETTY_PRINT);
```

The user should now be visible at http://localhost:4000/auth

### Todo

- [ ] Update Github Action to use the emulator in integration tests
- [ ] Integration tests work when using the auth emulator
- [ ] Try to not break backward compatibility (every project _should_ have a project ID available, but not every project has)
- [ ] Add docs
- [ ] Decide whether or not to test with the real Auth database once the emulator is fully integrated

:octocat:


MGatner:
For CI/CD I highly recommend using the `firebase emulators:exec` command. Here's my current example of it in action: https://github.com/tattersoftware/codeigniter4-firebase/blob/b5a0d370a86b47773c28bd8c896e17597d6191ec/.github/workflows/test.yml#L83

arendjantetteroo:
@jeromegamez we would be interested in having the emulator supported, what are your plans? 
What can we do to help? 
I assume this needs to be updated to support the 6.x branch, what else is missing? 

We currently work around this with a seperate factory object that only instantiates this library when we are in prod mode, and in dev mode work around it by assuming the emulator and doing unsigned verifies.

jeromegamez:
@arendjantetteroo Thank you for your interest! I have a working POC in a private GitHub Sponsors Repository, but due to lacking resources/headspace haven't had the time yet to make it releasable. At the moment, all URLs are hardcoded in the SDK (it's rooted from a time far before the Emulators existed), so a substantial overhaul is needed, and unfortunately, I can't give an ETA for it.

arendjantetteroo:
@jeromegamez how can i help?

PR #672, opened 2021-11-26
Remove shell prompt

Vladnev:
An extra character has been removed for easy copying of the composer command


jeromegamez:
Thanks!

PR #682, opened 2022-01-23
feat: add getTarget method to CloudMessage class

ankurk91:
Hi,

I am building a package around this package, where a developer can set target to `CloudMessage` and my package will be consuming the class instance, but there is no way to retrieve the target.

I have added a method to retrieve the target, this will be a nice addition since there is already a `hasTarget` method is the class.

```php
<?php

$message = CloudMessage::new()->withChangedTarget('topic', 'nice topic');

if($message->hasTarget()) {

 $target = $message->getTarget();
 // do something with target
 
}
```
Thanks.

### DIFF
--- a/src/Firebase/Messaging/CloudMessage.php
+++ b/src/Firebase/Messaging/CloudMessage.php
@@ -282,6 +282,11 @@ final class CloudMessage implements Message
         return (bool) $this->target;
     }
 
+    public function getTarget(): ?MessageTarget
+    {
+        return $this->target;
+    }
+
     /**
      * @return array<string, mixed>
      */
--- a/tests/Unit/Messaging/CloudMessageTest.php
+++ b/tests/Unit/Messaging/CloudMessageTest.php
@@ -164,4 +164,14 @@ final class CloudMessageTest extends TestCase
             ]],
         ];
     }
+
+    public function testGetTarget()
+    {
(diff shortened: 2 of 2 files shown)


MGatner:
This should be a pretty easy test to add. See [https://github.com/kreait/firebase-php/tree/6.x/tests/Unit/Messaging](https://github.com/kreait/firebase-php/tree/6.x/tests/Unit/Messaging).

ankurk91:
@MGatner 
Test case has been a added :crossed_fingers:

jeromegamez:
The CloudMessage not meant to be used outside of the SDK. If it was added to the class, it would also have to be added to the Message interface, and I'm not prepared to do that "just" to satisfy a requirement outside of the scope of the SDK, which doesn't need it. 

If I was to add this method, it would invite other requests to add getters for all the possible properties of a message, with the argument that this getter has been provided as well. I hope you can understand.

You could use the `jsonSerialize()` method (or, more safely, `json_encode()` + `json_decode()` the message) to then get the target out of it. An alternative would be to create your own implementation of the `Message` interface and wrap the CloudMessage inside it (e.g. with a `_call()` method to proxy the calls to the `CloudMessage`).

I'm not at the computer right now, but I can provide examples later if you need them.

ankurk91:
Thanks, i like your suggestion to create my own implementation.

PR #686, opened 2022-01-26
Add bulk remove option for RTDB references

davidblnc:
Hi!

This PR looks to add a more convenient way to remove data in bulk from a database. This is already possible by using `update` with an array with the paths as keys and null as values, which is what this PR does under the hood.

I added an optional argument to the `Reference::remove` method to do so, but maybe it would've been a better idea to create a different `Reference::bulkRemove` method. If you think so, I'm more than willing to change it to your liking.

Of course, feel free to edit anything (or request me to do so) or outright reject the PR if you think it doesn't fit.

Thanks for maintaining the library :)

### DIFF
--- a/docs/realtime-database.rst
+++ b/docs/realtime-database.rst
@@ -412,7 +412,15 @@ You can also delete by specifying null as the value for another write operation 
 
     $database->getReference('posts')->set(null);
 
-You can use this technique with ``update()`` to delete multiple children in a single API call.
+You can also delete in bulk passing an array of children to the function:
+
+.. code-block:: php
+
+    $data->getReference('posts')->remove([
+        'post1',
+        'post7',
+        'post15'
+    ]);
 
 *********************
 Database transactions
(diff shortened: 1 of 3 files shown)


jeromegamez:
This feature makes a lot of sense, thank you!

Only one thing: instead of changing the signature of the existing `remove()` method, I'd prefer an additional `removeChildren(array $keys)` method, I think that would be more explicit and would remove the necessity/"complexity" of the if/else.

davidblnc:
Done! I also changed the example code to show the use of nested paths.

jeromegamez:
Would you like to change the committer address (test@test.com) before I merge this? 🙈

davidblnc:
Wow! Thanks for letting me know, I wasn't aware this was public. Just changed it :)

jeromegamez:
Thank you for your contribution, it's much appreciated 🌺, especially that you added tests and documentation! 😍

PR #688, opened 2022-01-28
Support verifySessionCookie

arma7x:
Add support for verifySessionCookie


jeromegamez:
Thank you! I already started implementing it, including tests, but I will add you as a co-author, since you did the same work 😅

jeromegamez:
Here is the commit 31c05665189424cd5c4a76d7cee8d3eadedabcf6

PR #699, opened 2022-05-05
Fix JSON encoding of nested lists

cblavi-skitude:
Hi! 

I'm creating this PR because when using the JSON_FORCE_OBJECT param in json_encode function, it also casts to objects all arrays inside it. For example:

[1, 2, 3] will be translated to { '0': 1, '1': 2, '2': 3 }

I'm not sure that this behaviour is expected and in previous versions wasn't performing like this.

Kind regards!


jeromegamez:
Good catch, thank you!

PR #702, opened 2022-05-23
linkEmailProvider

WardVH:


jeromegamez:
Thank you for your contribution! Could you please restore the final keyword on the class and provide a PR description referencing the corresponding documentation?

jeromegamez:
I just did a test run with your proposed changes and it seems that what it does is updating the email of an existing user. You can achieve the same with calling

```php
$auth->updateUser(['email' => $newEmail]);
// or
$auth->updateUser(['verifiedEmail' => $newEmail]);
```

Feel free to add further comments if you disagree - perhaps I just haven't seen the difference…

PR #723, opened 2022-07-28
feat: set default http timeouts

ankurk91:
ref https://github.com/kreait/laravel-firebase/pull/146

Lets set sensible default timeouts to HTTP client.

I have taken these values from Laravel framework HTTP client.

This should not be a breaking change to existing users since firebase server is not going to wait for us indefinitely.


jeromegamez:
Merged with  0c10881452b94ea5b2b52ceb4bca27776dcc0d54, thank you! 🙏

PR #727, opened 2022-08-17
Add ability to get recently registered users and last active users

f-balazs:
Hi,

I added two new methods to get recently registered users and last active users both can be called with custom limit. I also added a new API call based on the documentation: https://cloud.google.com/identity-platform/docs/reference/rest/v1/projects.accounts/query

This is my first PR, thank you for your patient and kind feedback! ☺️


jeromegamez:
Thank you for making your first PR one for this project; it means a lot!

Querying the accounts with `accounts/query` would be a great addition to the SDK, much more convenient than the `listUsers()` method!

There are a couple of things I would do differently:

#### Methods

I wouldn't provide dedicated methods for recently registered and last active users - these are specific requirements that could change from project to project, so this is something I would consider to be the responsibility of the projects, not the SDK.

That's why I would move `queryUsers(...)` methods from the `ApiClient` to the `Auth` class and remove the `recentlyRegisteredUsers()` and `lastActiveUsers()` methods.

#### Method parameters

Instead of having single parameters for each query parameter (limit, offset, sortby, etc.) I would introduce a `UserQuery` value object that can also be constructed with an array. Here's an example how this could look like:

```php
# ❌
public function queryUsers( int $limit = 500, int $offset = 0, string $sortBy = null, string $order = null);

# 🚀
/**
 * @param UserQuery|array $query
 */
public function queryUsers($query): array
```

The query could the be built like this (example)

```php
$userQuery = UserQuery::all()
    ->sortedBy('CREATED_AT')
    ->inAscendingOrder()
    ->withOffset(1)
    ->limitedTo(25)

# or

$userQuery = UserQuery::fromArray([
    'sortBy' => 'CREATED_AT',
    'order' => 'ASC',
    'offset' => 1,
    'limit' => 25,
]);
```

In the method it could then look like this:

```php
/**
 * @param UserQuery|array $query
 */
public function queryUsers($query): array
{
    $query = is_array($query) ? UserQuery::fromArray($query) : $query;

    // ...
}
```

This would allow us to be more flexible if the number of possible parameters changes - for example, on https://cloud.google.com/identity-platform/docs/reference/rest/v1/projects.accounts/query#request-body there's also an `expression` field which we might want to implement later without having to change the method signature.

You could also have a look at https://github.com/kreait/firebase-php/blob/6.x/src/Firebase/Messaging/CloudMessage.php - although it's far too complex, I think it illustrates what I mean (ValueObject + PHPStan typed array parameters).

#### Next steps

I'm quite picky about how I think things should look 😅 and I realize it's a lot to ask because it would require many changes and iterations until this feature is released.

If you're willing to take on this journey (and the pain that comes with it 😂) you could start with the points I mentioned so far and I'll give out feedback until it's in a state where I would say "okay, the rest is just rebasing, adding tests and documentation, etc." and merge it.

If kickstarting the feature is "enough" for you (it definitely is for me 🙏), I can take it from here, and I will make sure that your contribution is immortalized in the git history 💪 

Either way, thank you very much for your contribution, this feature will make people happy, I'm sure!

f-balazs:
Thank you for your feedback! All of your points are make sense, and I tried to implement the modifications in the way you asked.

Question:
PHPStan returns with Method Kreait\Firebase\Auth\UserQuery::fromArray() has parameter $data with no value type specified in iterable type array. at the following lines, but I can't figure it our what is wrong:

```php
/**
 * @param array {
 *     limit?: positive-int,
 *     offset?: positive-int,
 *     sortBy?: SortByField|non-empty-string,
 *     order?: SortByOrder|non-empty-string
 * } $data
 */
public static function fromArray(array $data): self
```

jeromegamez:
(I edited your comment only to wrap the code snippet so that it is more readable)

The problem is not easy to spot, but PHPStan will recognize the shape definition when you change `@param array {` with `@param array{`, so, without a whitespace in-between.

Great progress, by the way, you're quick!

jeromegamez:
I really like that you put in the extra work to create value objects for the possible fields to ensure that only known/valid values can be specified, but I think that might be too much overhead and a limiting factor in case new fields are supported by Firebase.

I don't know if I mentioned this earlier somewhere, but I'm planning on removing validation logic and relying on static analysis as much as I can with the next major release.

Could you remove the value objects again, and instead add the constants into the `UserQuery` class?

```php
class UserQuery implements \JsonSerializable
{
    public const FIELD_USER_ID = 'USER_ID';
    public const FIELD_NAME = 'NAME';
    // ...
    
    public const SORT_ORDER_ASC = 'ASC';
    public const SORT_ORDER_DESC = 'DESC';
// ...
}
```

You could then write the shape of the `fromArray()` method like this:

```php
/**
 * @param array{
 *     limit?: positive-int,
 *     offset?: positive-int,
 *     sortBy?: self::FIELD_*,
 *     order?: self::SORT_ORDER_*
 * } $data
 */
public static function fromArray(array $data): self
```

MGatner:
What awesome teamwork.

f-balazs:
I removed the value objects.
Your example throws a PHPStan error, I used this comment instead:

```php
/**
     * @param array{
     *     limit?: positive-int,
     *     offset?: positive-int,
     *     sortBy?: value-of<self::VALID_SORT_BY_VALUES>,
     *     order?: value-of<self::VALID_SORT_BY_ORDERS>
     * } $data
     */

    public static function fromArray(array $data): self
```

jeromegamez:
The annotations I proposed work for me when I run `phpstan` on the command line, but if it doesn't for you, we can keep it that way and I'll take care of them later 🤞.

The runtime checks throwing an `InvalidArgumentException` are not necessary and can be removed (that's what I meant with relying on PHPStan earlier).

Something I didn't think about earlier: We can use `int<0, 500>` instead of `positive-int` to ensure a correct limit. I'm not sure the Firebase REST API support 0 as a limit, the integration tests should confirm that, we can take care of that later.

I also think that we should not pre-set the parameters but make them `null` by default and let the Firebase API decide what to do.

jeromegamez:
(Again, if I get on your nerves with my change requests, feel free to let me know and I'll take over - you did the most relevant work already and the rest is just syntactic sugar ☺️)

f-balazs:
> (Again, if I get on your nerves with my change requests, feel free to let me know and I'll take over - you did the most relevant work already and the rest is just syntactic sugar ☺️)

I did the suggested modifications. I know that writing modification requests is more time consumind than do it at this stage, so please, feel free to spice it your own taste. ☺️

I just wanted to do it well if I do something... Thank you for your patient!

Cheers! 🎉🍺

jeromegamez:
You did really well, in my opinion! Thank you so much! I will then continue working on it, including tests, docs, etc. and let you know once it's ready to be released.

Thank you so much!

jeromegamez:
This has now been merged and your contribution is attributed in 85dd7971033933193ad9301f81f22efa1e101569 🥳 

If you want to have a look at the final result of the UserQuery class, you can find it at https://github.com/kreait/firebase-php/blob/6.x/src/Firebase/Auth/UserQuery.php

Thanks again for you valuable contribution, I appreciate it a lot! 🙌

PR #736, opened 2022-09-26
added sound support in default notification

guibothh:
Added sound support on default notification


jeromegamez:
Thank you for taking the time to contribute to this project.

According to https://firebase.google.com/docs/reference/fcm/rest/v1/projects.messages#Notification , the notification payload doesn't support/use a `sound` key.

If you're sure it does, could you please provide me with a source?

Please note that the SDK uses the HTTP v1 API, not the legacy API.

PR #747, opened 2022-11-04
feat(app-check): added support for firebase app check

### LINKED ISSUE #693: Does this package support Firebase app check?
For verifying firebase app check token?

[https://firebase.google.com/docs/app-check/custom-resource-backend](https://firebase.google.com/docs/app-check/custom-resource-backend)

nhaynes:
# Description

I added support for **Firebase App Check** so the SDK can be used to implement a [custom provider](https://firebase.google.com/docs/app-check/custom-provider) and can also be used to [verify app check tokens](https://firebase.google.com/docs/app-check/custom-resource-backend). 

I wanted to use App Check to protect custom backend resources for a project of mine and was frustrated to find out that it's only available in the official Node.JS SDK.

Closes [#693](https://github.com/kreait/firebase-php/issues/693)

# Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have made corresponding changes to the documentation


jeromegamez:
Wow, this looks like you have put a lot of work into this, and I appreciate that you stuck with the code style of the SDK (for better or worse, it's an old code base, after all, 😅), and even docs and tests! That's the best PR this project has received so far, I love it!

It will not be long before I drop support for PHP <8.1 (I changed my support policy to only support actively supported PHP versions), so this feature will certainly be added to the upcoming 7.0 release. I'm giving you this heads-up because it means a rebase will be in your future 😅.

This could also be a good start for deprecating https://github.com/kreait/firebase-tokens-php and incorporating Custom Token generation and ID Token verification using `firebase/php-jwt` into the SDK, but that's another topic (that could also lead to another rebase).

I appreciate what you did already - should I makle a lot of change requests and comments (I only quickly skimmed the code, so I don't know if there will be many 😄), please keep in mind that I only have the best intentions and appreciate the time and work you're investing 🙏 

I'm pretty swamped at the moment, but I'll try to look at the details as soon as possible! In the meantime, I'll enable running the workflows, in case it already brings something to light that you want/need to address 💪

jeromegamez:
Ha, the tests are already failing due to the PHP requirements. Could you change the requirement in the composer.json to `"php": "~8.1.0|~8.2.0"`, then do a `composer bump` while we're at it, and update the requirements in the `.github/workflows` directory accordingly?

nhaynes:
@jeromegamez Thanks for the feedback! Adding it to 7.x seems like a sensible plan to me. I'll work on getting the php requirements sorted out for now.

jeromegamez:
@nhaynes I will create a new 7.x branch with the updated PHP requirements so that you can re-target the PR.

I will now have a closer look at the changes, but from what I can see, I can already say that I would prefer not to have the phpfastcache dependency (= implementation) - there are already the `withVerifierCache()` and `withAuthTokenCache()` methods in the factory, and I see the App Check Token caching either using the same cache as provided to `withVerifierCache()` or an additional `withAppCheckTokenVerifierCache()`.

This way, we wouldn't have an additional dependency, and the project implementing the SDK can inject its own cache implementation - for example, in Laravel https://github.com/kreait/laravel-firebase/blob/14812145bbd8598ec2c9f447aac166b451cfadd6/src/FirebaseProjectManager.php#L94-L105

composer.json
```diff
@@ -38,6 +38,7 @@
         "kreait/firebase-tokens": "^3.0",
         "lcobucci/jwt": "^4.1",
         "mtdowling/jmespath.php": "^2.6.1",
+        "phpfastcache/phpfastcache": "^9.1",
```
jeromegamez: See https://github.com/kreait/firebase-php/pull/747#issuecomment-1304133629 for why I would like to have phpfastcache removed again 😅 

Could you instead please add `firebase/php-jwt` as a direct dependency? At the moment it's transitively included through the Google Auth library, but if they were to remove it, we wouldn't have it anymore either.

src/Firebase/AppCheck.php
```diff
@@ -0,0 +1,53 @@
...
+
+    public function __construct(
+        ApiClient $client,
+        AppCheckTokenGenerator $tokenGenerator,
+        AppCheckTokenVerifier $tokenVerifier
+    ) {
+        $this->client = $client;
+        $this->tokenGenerator = $tokenGenerator;
+        $this->tokenVerifier = $tokenVerifier;
+    }
```
jeromegamez: Just a note, you can resolve it instantly :D - it will be so satisfying to replace this with once support for PHP <8.1 is dropped 🤤 

```php
final class AppCheck implements Contract\AppCheck
{
    public function (
        private readonly ApiClient $client,
        private AppCheckTokenGenerator $tokenGenerator,
        private AppCheckTokenVerifier $tokenVerifier,
) {
    // …
}
```

src/Firebase/AppCheck/AppCheckTokenOptions.php
```diff
@@ -0,0 +1,62 @@
...
+{
+    private ?string $ttl;
+
+    private function __construct(?string $ttl = null)
+    {
+        $this->ttl = $ttl;
+    }
+
+    /**
+     * @param array<string, mixed> $data
```
jeromegamez:
```suggestion
     * @param AppCheckTokenOptionsShape $data
```

src/Firebase/AppCheck/AppCheckTokenOptions.php
```diff
@@ -0,0 +1,62 @@
...
+     */
+    public static function fromArray(array $data): self
+    {
+        if (!array_key_exists('ttl', $data)) {
+            throw new InvalidArgumentException('The "app_id" key is missing from the token data.');
+        }
```
jeromegamez: `ttl` instead of `app_id` in the error message

I'm thinking about removing many of these runtime checks, not just here but in many places of the SDK where PHPStan/Psalm PHPDocs are in place.

For example, here, the `AppCheckTokenOptionsShape` already defines that a `ttl` key _must_ be present, so I'm wondering if runtime checks are even necessary.

So, unless you have strong opinions against it, I think it would be good to remove runtime checks where static analysis would already catch the problem.

(Strong Opinions against it: except "But most people don't use Static Analysis Tools" - because they should™ 😅)

src/Firebase/AppCheck/AppCheckTokenVerifier.php
```diff
@@ -0,0 +1,90 @@
...
+/**
+ * @internal
+ */
+class AppCheckTokenVerifier
+{
+    private const APP_CHECK_ISSUER_PREFIX = 'https://firebaseappcheck.googleapis.com/';
+    private string $projectId;
+    private CachedKeySet $keySet;
+
+    public function __construct(string $projectId, CachedKeySet $keySet)
```
jeromegamez:
```suggestion
    /**
     * @param non-empty-string $projectId
     */
    public function __construct(string $projectId, CachedKeySet $keySet)
```

Annotations like this in general could be added or improved, but don't invest too much time into it, just when you are working at those locations anyway 🙏

src/Firebase/AppCheck/DecodedAppCheckToken.php
```diff
@@ -0,0 +1,146 @@
...
+        $this->app_id = $app_id;
+        $this->aud = $aud;
+        $this->exp = $exp;
+        $this->iat = $iat;
+        $this->iss = $iss;
+        $this->sub = $sub;
+    }
+
+    /**
+     * @param array<string, mixed> $data
```
jeromegamez:
```suggestion
     * @param DecodedAppCheckTokenShape $data
```

This would remove the need for the `array_key_exists` checks - it's a public function, but in reality, this will only be used by the SDK itself.

src/Firebase/AppCheck/AppCheckTokenVerifier.php
```diff
@@ -0,0 +1,90 @@
...
+    /**
+     * @param string $token the Firebase App Check JWT token to decode
+     *
+     * @throws FailedToVerifyAppCheckToken if the token could not be verified
+     * @throws InvalidAppCheckToken if the token is invalid
+     */
+    private function decodeJwt(string $token): DecodedAppCheckToken
+    {
+        try {
+            $payload = (array) JWT::decode($token, $this->keySet);
```
jeromegamez, line 63: I'm not 100% sure, but I think we could do a

```php
/** @var DecodedAppCheckTokenShape $payload */
$payload = (array) JWT::decode($token, $this->keySet);
```

here, together with my comment on the `DecodedAppCheckToken` this should save us some code 🤞

src/Firebase/Factory.php
```diff
@@ -249,6 +252,35 @@ public function withClock(object $clock): self
...
+
+        $http = $this->createApiClient([
+            'base_uri' => 'https://firebaseappcheck.googleapis.com/v1/projects/'.$projectId,
+        ]);
+
+        $keySet = new CachedKeySet(
+            'https://firebase.google.com/docs/reference/appcheck/rest/v1/jwks#resource:-publicjwkset',
+            new Client(),
+            new HttpFactory(),
+            CacheManager::getInstance('files'),
```
jeromegamez: This is what I meant when I said that I'd prefer to use an injected instance of a cache instead of declaring it here ourselves - a file cache is undoubtedly good enough for most installations with a filesystem, but there are environments and/or preferences that would require another implementation

jeromegamez (changes requested):
The PR looks very good overall, everything I noted (except phpfastcache) are just minor things that don't really make a change in functionality, so if this is too cumbersome (and you already invested quite some time), I can do it later.

I will create the new 7.x branch with the updated PHP requirements within the hour so that you can retarget the PR (as far as I know this can be done without creating a new PR, but I haven't done that myself yet 😅)

After the next iteration, I'll check out the code on my computer/IDE, that's when I usually find what I missed earlier. :D

jeromegamez:
https://github.com/kreait/firebase-php/tree/7.x is the new default branch 🥳

jeromegamez:
I updated the `beste/*` dependencies to support PHP 8.1 and 8.2, so they should not be the reason for failing tests anymore. Also: #748  - I don't know what the best approach would be here, but we'll figure something out 🤞

jeromegamez:
Sorry for the comment spam and for the back and forth 😅.

Once you've removed the PHPFastCache dependency, and a short review in my IDE, I'll merge the PR.

Language Level migrations and such is positively none of your business 😄

jeromegamez:
#748 is now merged into the 7.x branch.

src/Firebase/AppCheck/AppCheckTokenVerifier.php
nhaynes: I'm not sure I follow this one. Are you saying the type-hint is all we need here and we shouldn't concern ourselves with the exception handling? I'm not sure I understand how static analysis is going to help us if the developer provides a string that isn't a valid token at runtime.

nhaynes: Nevermind, I see what you mean now. You weren't referring to the exception handling, you referring to the `fromArray` call and the type hint there. We are assuming that if the JWT parses correctly, then the fields we expect to be there should be there.

jeromegamez:
Thank you for addressing the things I wanted to have changed, really appreciate it!

The integration tests in this branch run without errors, but I just noticed they do because the Secrets are not available in PR GitHub actions, so they're all skipped.

On my local computer, the unit tests all succeed, but the Integration Tests don't, probably because I don't have a valid AppCheck token

```
GuzzleHttp\Exception\ClientException: Client error: `POST https://firebaseappcheck.googleapis.com/v1/projects/apps/com.example.test-app` resulted in a `404 Not Found` response
```

As I haven't worked with AppCheck Tokens myself yet, could you tell me how to generate one / what I have to do to be able to generate one? The Firebase project I test with has no associated apps, and unfortunately, I'm also not an app developer, so I wouldn't know on my own how to do it 😅 

Other than that, the PR looks good (as the initial one, just better :D), some minor things that I can do myself just before or after merging it.

There's one other thing concerning the AppCheckTokenOptions, but I will comment on that in the code.

I feel like I'm saying "Thank you" too often, but it's heartfelt so I will keep doing it :D

src/Firebase/AppCheck/AppCheckTokenOptions.php
```diff
@@ -0,0 +1,72 @@
...
+
+        if (null === $ttl) {
+            return new self();
+        }
+
+        if (!is_numeric($ttl)) {
+            throw new InvalidAppCheckTokenOptions('The ttl must be a number.');
+        }
+
+        if ($ttl < 1800 || $ttl > 604800) {
```
jeromegamez, line 34: I don't know why PHPStan didn't catch this, but the AppCheckTokenOptionsShape defines `ttl` to be `string|null`, but here, the check is done with `int`s.

Locally on my computer this gives me 

```
2) Kreait\Firebase\Tests\Integration\AppCheckTest::testCreateTokenWithCustomTtl
TypeError: Kreait\Firebase\AppCheck\AppCheckTokenOptions::__construct(): Argument #1 ($ttl) must be of type ?string, int given, called in /Users/jg/Code/kreait/firebase-php/src/Firebase/AppCheck/AppCheckTokenOptions.php on line 47
```

Not sure what the best approach here could be. I know the Firebase API expects it as a string, so perhaps the object could accept `int`, but expose `string`?

src/Firebase/AppCheck/AppCheckTokenOptions.php
```diff
@@ -0,0 +1,72 @@
...
+        ];
+    }
+
+    /**
+     * @return AppCheckTokenOptionsShape
+     */
+    public function jsonSerialize(): array
+    {
+        return $this->toArray();
+    }
```
jeromegamez: Could you please check if the Options are actually JSON-serialized somewhere and/or the `toArray()` method is used anywhere? If not, or if callers could be rewritten to not use json_decode/toArray(), we could perhaps remove them.

(I added those and `__toString()` methods almost everywhere in the beginning and now I regret it because I have to keep them for BC, but also test them :D)

src/Firebase/AppCheck/AppCheckTokenVerifier.php
```diff
@@ -0,0 +1,96 @@
...
+     * @param non-empty-string $projectId
+     */
+    public function __construct(string $projectId, CachedKeySet $keySet)
+    {
+        $this->projectId = $projectId;
+        $this->keySet = $keySet;
+    }
+
+    /**
+     * Verfies the format and signature of a Firebase App Check token.
```
jeromegamez:
```suggestion
     * Verifies the format and signature of a Firebase App Check token.
```

When I start to get nitpicky, it's usually a good sign 😂

src/Firebase/AppCheck/AppCheckTokenGenerator.php
```diff
@@ -0,0 +1,53 @@
...
+    public function createCustomToken(string $appId, ?AppCheckTokenOptions $options = null): string
+    {
+        $now = $this->clock->now()->getTimestamp();
+        $payload = [
+            'iss' => $this->serviceAccount->getClientEmail(),
+            'sub' => $this->serviceAccount->getClientEmail(),
+            'app_id' => $appId,
+            'aud' => self::APP_CHECK_AUDIENCE,
+            'iat' => $now,
+            'exp' => $now + 300,
```
jeromegamez, line 44: Is 300 a configurable value? How does it relate to the `ttl`? Could it be `$exp = $options->ttl ?? 300` or something?

src/Firebase/AppCheck/DecodedAppCheckToken.php
```diff
@@ -0,0 +1,119 @@
...
+    private string $exp;
+    private string $iat;
+    private string $iss;
+    private string $sub;
+
+    /**
+     * @param array<string> $aud
+     */
+    private function __construct(
+        string $app_id,
```
jeromegamez: Since we have PHP 8.1 as the minimum version, we could go with

```php
private function __construct(
    /** @var non-empty-string $app_id */
    public readonly string $app_id,
    /** @var array<string> $aud /*
    public readonly array $aud,
    public readonly string $exp,
    // and so on
) {
}
```

This way, we could get rid of the property declarations above and the getter functions below 🚀

jeromegamez (changes requested):
I noticed a few more things - as before, nothing serious, but for your consideration 🤞

jeromegamez:
I found out about AppCheck debug tokens and will try it out in the next few days. Once I get it working, I'll merge your PR and get it ready for the new SDK version (i've not decided yet if it will be 7.0 or 7.1, but it will come 😊)

src/Firebase/AppCheck/ApiClient.php
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    /**
+     * @throws AppCheckException
+     *
+     * @return AppCheckTokenShape
+     */
+    public function exchangeCustomToken(string $appId, string $customToken): array
+    {
+        $response = $this->requestApi('POST', 'apps/'.$appId, [
```
jeromegamez: This request results in a 404 - according to the [documentation](https://firebase.google.com/docs/reference/appcheck/rest/v1/projects.apps/exchangeCustomToken#http-request), the target URL should be `apps/<appId>:exchangeCustomToken`, how did you make this work in your tests?

src/Firebase/Factory.php
```diff
@@ -245,6 +257,39 @@ public function withClock(object $clock): self
...
+        if ($projectId === '') {
+            throw new RuntimeException('Unable to determine the Firebase Project ID');
+        }
+
+        if ($serviceAccount === null) {
+            throw new RuntimeException('Unable to use AppCheck without credentials');
+        }
+
+        $http = $this->createApiClient([
+            'base_uri' => 'https://firebaseappcheck.googleapis.com/v1/projects/'.$projectId,
```
jeromegamez:
```suggestion
            'base_uri' => 'https://firebaseappcheck.googleapis.com/v1/projects/'.$projectId.'/',
```

The base URI must end with a slash. Otherwise, a

```php
$http->get('GET', 'app/id')
```

will request `https://firebaseappcheck.googleapis.com/v1/projects/app/id` (which it did for me when I debugged the PR)

src/Firebase/Factory.php
```diff
@@ -245,6 +257,39 @@ public function withClock(object $clock): self
...
+        ]);
+
+        $keySet = new CachedKeySet(
+            'https://firebase.google.com/docs/reference/appcheck/rest/v1/jwks#resource:-publicjwkset',
```
jeromegamez: Why is this a link to the Firebase documentation? 😅 

```suggestion
            'https://firebaseappcheck.googleapis.com/v1/jwks',
```

tests/Integration/AppCheckTest.php
```diff
@@ -0,0 +1,61 @@
...
+ */
+final class AppCheckTest extends IntegrationTestCase
+{
+    public AppCheck $appCheck;
+    public string $appId;
+
+    protected function setUp(): void
+    {
+        $this->appCheck = self::$factory->createAppCheck();
+        $this->appId = 'com.example.test-app';
```
jeromegamez: How did you make this work when running the integration tests?

tests/Integration/AppCheckTest.php
```diff
@@ -0,0 +1,61 @@
...
+
+    public function testCreateTokenWithCustomTtl(): void
+    {
+        $token = $this->appCheck->createToken($this->appId, [
+            'ttl' => 1800,
+        ]);
+
+        $this->assertInstanceOf(AppCheckToken::class, $token);
+        $this->assertIsString($token->token());
+        $this->assertIsNumeric($token->ttl());
```
jeromegamez: The return types don't need to be checked, the `AppCheckToken` class already ensures that. Same in other tests

jeromegamez (changes requested):
I should have tested locally first instead of just skimming over the Github Web page 😅

src/Firebase/AppCheck/AppCheckTokenGenerator.php
nhaynes: I based this off of what the official node package is doing. They are always setting the expiration for the custom token to 300. 

I think because it creates a custom token and then immediately exchanges the custom token for a App Check token, that the time here doesn't matter much. I'm guessing they chose 5 mins because it's long enough to ensure a race condition doesn't occur here with the token exchange but it's short enough that it is a short lived token?

tests/Integration/AppCheckTest.php
nhaynes: Apparently I didn't have them running. I didn't realize the integration tests were being auto-skipped locally.

I've now created a test project and added the necessary `test_credentials.json` and `test_rtdb.json` files and some of the integration tests are passing but I'm not sure what I need to put in `test_devices.json` to get the rest passing? What exactly are the registration tokens here?

jeromegamez: If you run only the AppCheck tests (`phpunit tests/Integration/AppCheckTest.php`), you might not need the FCM registration tokens. I generate one with https://github.com/firebase/quickstart-js/tree/master/messaging and then copy it as a JSON array into the `test_devices.json` file (`["<device token>"]`).

That's a good reminder to add a `CONTRIBUTING.md` 😅

jeromegamez:
@nhaynes I think it should be possible to prevent most of the PHPStan errors when you run `composer install-tools` and then `composer tests`. Also, could you re-request a review (in the right sidebar at the top) once you pushed something, so that I can get a notification and don't have to let you wait? 🙏 

Just because I noticed the 

```php
public function __construct(
        private ServiceAccount $serviceAccount,
        private ClockInterface $clock = SystemClock::create(),
) {
}
```

src/Firebase/AppCheck/AppCheckTokenGenerator.php
jeromegamez: Sounds reasonable, thanks for clarifying! 🙌

src/Firebase/AppCheck/ApiClient.php
```diff
@@ -0,0 +1,68 @@
...
+use Psr\Http\Message\UriInterface;
+use Throwable;
+
+/**
+ * @internal
+ *
+ * @phpstan-type AppCheckTokenShape array{
+ *     token: string,
+ *     ttl: string
+ * }
```
jeromegamez: I think this would be better located in the `AppCheckToken` class.

src/Firebase/AppCheck/AppCheckToken.php
```diff
@@ -0,0 +1,70 @@
...
+        ];
+    }
+
+    /**
+     * @return AppCheckTokenShape
+     */
+    public function jsonSerialize(): array
+    {
+        return $this->toArray();
+    }
```
jeromegamez: I know I added `toArray()` and `jsonSerialize()` to value objects in the past as well. Still, I think as long as the `AppCheckToken` is not serialized in another SDK component, we should omit them here. If a project want to serialize the app check tokens, they can do it themselves 😅

src/Firebase/AppCheck/AppCheckTokenGenerator.php
```diff
@@ -0,0 +1,51 @@
...
+ */
+class AppCheckTokenGenerator
+{
+    private const APP_CHECK_AUDIENCE = 'https://firebaseappcheck.googleapis.com/google.firebase.appcheck.v1.TokenExchangeService';
+
+    public function __construct(
+        private ServiceAccount $serviceAccount,
+        private ClockInterface $clock = SystemClock::create(),
+    ) {
+    }
```
jeromegamez:
```suggestion
    
    private ClockInterface $clock;

    public function __construct(
        private ServiceAccount $serviceAccount,
        ?ClockInterface $clock = null,
    ) {
        $this->clock = $clock ?? SystemClock::create();
    }
```

src/Firebase/AppCheck/VerifyAppCheckTokenResponse.php
```diff
@@ -0,0 +1,58 @@
...
+ *
+ * @phpstan-type VerifyAppCheckTokenResponseShape array{
+ *     appId: non-empty-string,
+ *     token: DecodedAppCheckTokenShape,
+ * }
+ */
+final class VerifyAppCheckTokenResponse implements JsonSerializable
+{
+    /**
+     * @param non-empty string $appId
```
jeromegamez:
```suggestion
     * @param non-empty-string $appId
```

src/Firebase/AppCheck/VerifyAppCheckTokenResponse.php
```diff
@@ -0,0 +1,58 @@
...
+     */
+    public function __construct(
+        private string $appId,
+        private DecodedAppCheckToken $token,
+    ) {
+        $this->appId = $appId;
+        $this->token = $token;
+    }
```
jeromegamez:
```suggestion
    public function __construct(
        public readonly string $appId,
        public readonly DecodedAppCheckToken $token,
    ) {
    }
```

This would make the `appId()` and `token()` getter functions obsolete. 

When using `private/public <type> $variable` as constructor parameter, you don't need to set them in the constructor body anymore.

When a property is only set in the constructor, we can always add the `readonly` qualifier 💪

src/Firebase/AppCheck/DecodedAppCheckToken.php
```diff
@@ -0,0 +1,106 @@
...
+     */
+    private function __construct(
+        private string $app_id,
+        private array $aud,
+        private string $exp,
+        private string $iat,
+        private string $iss,
+        private string $sub,
+    ) {
+    }
```
jeromegamez:
```suggestion
    private function __construct(
        public readonly string $app_id,
        /** @var array<string> $aud */
        public readonly array $aud,
        public readonly string $exp,
        public readonly string $iat,
        public readonly string $iss,
        public readonly string $sub,
    ) {
    }
```

Like the others above, this would help us get rid of all the related getter methods.

src/Firebase/AppCheck/AppCheckTokenOptions.php
```diff
@@ -0,0 +1,64 @@
...
+declare(strict_types=1);
+
+namespace Kreait\Firebase\AppCheck;
+
+use JsonSerializable;
+use Kreait\Firebase\Exception\AppCheck\InvalidAppCheckTokenOptions;
+
+/**
+ * @phpstan-type AppCheckTokenOptionsShape array{
+ *     ttl int|null,
```
jeromegamez:
```suggestion
 *     ttl?: int|null,
```

jeromegamez (changes requested):

nhaynes:
@jeromegamez I believe I've made all the improvements you requested above and I have successfully gotten the **Integration tests** for **AppCheck** to succeed locally (along with **Unit** and **PHPStan**). 

I started a `CONTRIBUTING.md` file to document the steps I had to take to get the integration tests to run locally. I managed to get most of them passing locally following these steps but I am still getting a few errors on the **Auth**, **DynamicLink**, and **RemoveConfig** modules. I'll dig into those later and update `CONTRIBUTING.md` if I find a solution.

nhaynes:
@jeromegamez I added the `guzzlehttp/psr7` library as a direct dependency to ensure the `HttpFactory` interface is available, should address those failed tests.

nhaynes:
@jeromegamez looks like that resolved the test issues and now it's just the code coverage blocking us. I'll add some additional tests this weekend.

jeromegamez:
@nhaynes But don't spend too much time on it, you've already invested so much of it!

I'll probably get to a final review on the week-end.

jeromegamez:
What baffles me a little is that the psr7 library had to be added directly - it should come with Guzzle via the Google auth library. But it doesn't hurt either, I'm just curious.

docs/app-check.rst
```diff
@@ -0,0 +1,79 @@
+##############
+App Check
+##############
```
jeromegamez, line 3:
```suggestion
########
App Check
########
```

docs/app-check.rst
```diff
@@ -0,0 +1,79 @@
...
+
+Before you start, please read about Firebase App Check in the official documentation:
+
+* `Introduction to the Admin Database API <https://firebase.google.com/docs/app-check>`_
+* `Verify App Check tokens from a custom backend <https://firebase.google.com/docs/app-check/custom-resource-backend>`_
+* `Implement a custom App Check provider <https://firebase.google.com/docs/app-check/custom-provider>`_
+
+*******************************
+Initializing the App Check component
+*******************************
```
jeromegamez, line 15:
```suggestion
************************************
Initializing the App Check component
************************************
```

docs/app-check.rst
```diff
@@ -0,0 +1,79 @@
...
+
+.. code-block:: php
+
+    use Kreait\Firebase\Exception\AppCheck\FailedToVerifyAppCheckToken;
+
+    $token = getallheaders()['X-Firebase-AppCheck'];
+
+    try {
+        $appCheck->verifyToken($token);
+    } catch (FailedToVerifyAppCheckToken $e) {
```
jeromegamez, line 62: I would document this the same way as, for example, the ID Token verification. Depending on the project, people might retrieve the AppCheck Token in different ways, `getallheaders()` is an implementation details which some users might interpret as "the thing to do" and then open issues when things don't work out 😅 

```suggestion
    $appCheckTokenString = '...';

    try {
        $appCheck->verifyToken($appCheckTokenString);
    } catch (FailedToVerifyAppCheckToken $e) {
```

docs/app-check.rst
```diff
@@ -0,0 +1,79 @@
+##############
+App Check
+##############
+
+The Firebase Admin SDK for PHP provides an API for verifying custom backends using Firebase App Check.
+
+Before you start, please read about Firebase App Check in the official documentation:
+
+* `Introduction to the Admin Database API <https://firebase.google.com/docs/app-check>`_
```
jeromegamez, line 9:
```suggestion
* `Introduction to Firebase App Check <https://firebase.google.com/docs/app-check>`_
```

docs/app-check.rst
```diff
@@ -0,0 +1,79 @@
...
+* `Verify App Check tokens from a custom backend <https://firebase.google.com/docs/app-check/custom-resource-backend>`_
```
jeromegamez, line 10:
```suggestion
* `Verify App Check tokens from a custom backend (Client-side) <https://firebase.google.com/docs/app-check/custom-resource-backend>`_
```

CHANGELOG-7.x.md
```diff
@@ -6,6 +6,11 @@ The most notable change is that you need PHP 8.1/8.2 to use the new version. The
 the SDK has introduced some breaking changes concerning parameter types, which should not affect your
 project in most cases.
 
+* Dropped support for not actively supported PHP versions. Supported PHP versions are 8.1.x and 8.2.x
+* Added direct dependency on `firebase/jwt`.
```
jeromegamez, line 10: I wouldn't add dependency changes to the changelog. The firebase-jwt library was already included via the Google Auth Library anyway 😅 
```suggestion
```

CHANGELOG-7.x.md
```diff
@@ -6,6 +6,11 @@ The most notable change is that you need PHP 8.1/8.2 to use the new version. The
...
+* Added support for verifying Firebase App Check Tokens. ([#693](https://github.com/kreait/firebase-php/issues/693))
+* Added support for generating Firebase App Check Tokens for a custom provider.
+
```
jeromegamez, line 13:
```suggestion
* Added support for generating and verifying Firebase App Check Tokens ([#747](https://github.com/kreait/firebase-php/issues/747))
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,31 @@
+# Contributing | Firebase Admin PHP SDK
```
jeromegamez, line 1:
```suggestion
# Contributing
```

jeromegamez (changes requested):
These are only doc-related changes, and I think we're ready then.

Could you also add `app-check` to the [docs/index.rst](https://github.com/kreait/firebase-php/blob/7.x/docs/index.rst) in the user guide section below the `user-management` entry?

jeromegamez:
Please don't work on the PR anymore, I already have it ready to be merged locally. I will first release a 7.0 and then a 7.1 with the AppCheck component shortly after 🤞 

Thank you so much for your work, I really, really appreciate it - and a lot of other SDK users will do the same, it's been brought up again and again :D 

If you haven't provided this in the context of a job, could you please enable GitHub sponsors or give me a link to a Buy me a coffee or something similar? I think you have definitely earned a token of gratitude 🙏

jeromegamez:
Thank you for your patience! I merged your PR with d2dd8d5a43ee179ab397982567234b39f5f478bd, ed58b866a428684df9cda73b110bf4ae7beaddeb, and 105a128a858e4ee6bf5b8cd72e7f19f8e8c14f1f - the changes I made  were only to adapt to the changes in the other parts of the SDK, but it's still all attributed to you ☺️ 

Thanks again for your work! It will be included in the 7.0 release, which shouldn't take too long. 🤞

nhaynes:
@jeromegamez awesome! thanks for finishing it up!

jeromegamez:
@nhaynes The component has been released with version 7.0 of the SDK. Again, thank you so much for adding it! 🙏 🌺

nhaynes:
@jeromegamez awesome! Great to see! Thanks for getting it across the finish line!

PR #748, opened 2022-11-04
PHP 8.1 Language Migration

jeromegamez:
:octocat:


composer.json
```diff
@@ -49,6 +49,7 @@
         "phpstan/phpstan": "^1.9.1",
         "phpstan/phpstan-phpunit": "^1.2.2",
         "phpunit/phpunit": "^9.5.26",
+        "roave/backward-compatibility-check": "^7.4",
```
MGatner, line 52: I love this library!

src/Firebase/Contract/Auth.php
```diff
@@ -47,7 +47,7 @@ interface Auth
     public function getUser($uid): UserRecord;
 
     /**
-     * @param array<Stringable|string> $uids
+     * @param non-empty-list<Stringable|string> $uids
```
MGatner, line 50: You might not care (I don't usually vendor tag), but this one is unique to PHPStan:
```suggestion
     * @param array<Stringable|string> $uids
     * @phpstan-param non-empty-list<Stringable|string> $uids
```

src/Firebase/RemoteConfig/ApiClient.php
```diff
@@ -22,10 +21,12 @@ class ApiClient
 {
     private ClientInterface $client;
     private RemoteConfigApiExceptionConverter $errorHandler;
+    private string $baseUri;
 
-    public function __construct(ClientInterface $client)
+    public function __construct(string $projectId, ClientInterface $client)
```
MGatner, line 26: Is BC check okay with this because of the `@internal`?

src/Firebase/Contract/Auth.php
jeromegamez: I never really thought about it and assumed PHPStan and Psalm supported what the other provides 🙈. Thank you for telling me, I think I will then just use `array` instead of `list`, I think vendor-prefixing would add to much overhead (for me 😅)

src/Firebase/RemoteConfig/ApiClient.php
jeromegamez: I thought about that, too, and I told myself "yes" because people shouldn't use internal classes directly.

There are other breaking changes, for example using `string` instead of `Stringable|string`, but it will be a major release after all, and I assume that most projects use simple strings anyway. I introduced `Stringable` back then because I wanted to move away from Value Object parameters (like `Uid`) without breaking projects.

My goal is being able to say in the release notes: "If you're not extending classes from the SDK and use the SDK as documented, this should be a drop-in replacement for the latest 6.x version". I hope that holds up in the end 😅

jeromegamez: Oh, and one other thing: I want to move away from runtime checks to static analysis checks, what are your thoughts about this?

src/Firebase/Contract/Auth.php
MGatner: Yeah that's where I am at with my own stuff. We had this conversation recently at CodeIgniter because we want to support a variety of tools people use downstream, especially IDEs. So we are using vendor tags. 

I think Psalm supports `non-empty-list` already but PHPStorm does not? It's been a bit since I looked. What I really want is a big spreadsheet of all the annotations and which tools support them 🙄

src/Firebase/RemoteConfig/ApiClient.php
MGatner: > I want to move away from runtime checks to static analysis checks, what are your thoughts about this?

100%! Runtime checks still can't even verify `string[]` or generics 😂 We would lose so much! I'm at the point where I consider my docblocks to be part of my code, and my static analysis tools "compile" all the code into PHP for runtime. Working with legacy stuff I end up with a fair amount of ignores or baselines, but my eventual goal is to have everything at PHPStan level 5 and then consider moving higher.

src/Firebase/Contract/Auth.php
jeromegamez: I'm using the PHPStan integration in PHPStorm and disabled the standard PHPStorm inspections that conflict with them.

Thinking about it, I think I will still keep the PHPStan-known annotations - I'm sure PHPStan and Psalm eventually will keep up with each other, and if they don't I hope people using Psalm will notify me and then I can add them as vendor-prefixed annotations 🤔

I might of course change my mind again, as I often do 😄

src/Firebase/RemoteConfig/ApiClient.php
jeromegamez: That sounds good! I'm not sure if I was clear with what I said or if we mean the same thing (we probably do), but I will spell it out just in case, and for others that my follow our conversation:

At the moment I'm doing runtime checks to "protect" users from "doing it wrong", for example providing invalid arguments like empty strings. By removing the runtime checks, they aren't protected anymore except if they use PHPStan/Psalm themselves in their projects. This would reduce the lines of code drastically in the messaging and user management components and perhaps motivate people not using static analysis to do so.

src/Firebase/Contract/Auth.php
MGatner: This is basically where I am at with my own OSS stuff, so no complaints here! 🤷‍♂️

src/Firebase/RemoteConfig/ApiClient.php
MGatner: Understood! I think I am 90% there. Maybe this goes back to our other thread, but I use a lot of PHPStan-specific stuff that I do t always assume is enforced at the user level. So if I believe my annotation is general enough to show up in a modern IDE then I won't recheck it:
```php
/**
 * @param array<string> $names
 */
function length(array $names): int {
    return strlen(reset($names));
}
```
(Ignoring the fact that this should have been variadic)

But if it is something I expect the user to do yet I am only able to enforce it in PHPStan I typically still use an Assert:
```php
/**
 * @param class-string<Clock>
 */
function length(string $handler): int {
    Assert::instanceOf($handler, Clock::class);

    return (new $handler)->now();
}
```

PR #760, opened 2023-01-23
Attach errors to NotFound exception to be able to always have the full error response available

pelletiermaxime:
Hello, I was a bit confused when looking at the NotFound Exception because the errors() method always returned an empty array. After a bit of research I found that you catched that specific exception in Messaging.php and created a new NotFound, but lost the errors array that was previously created by convertResponse(). A simple fix is to add back the errors  to becauseTokenNotFound().
I only tested this in 5.x because of limitations (php 7.4 and Guzzle 6), but the problem seems to also be in 7.x.

### DIFF
--- a/src/Firebase/Messaging.php
+++ b/src/Firebase/Messaging.php
@@ -58,7 +58,7 @@ final class Messaging implements Contract\Messaging
             $token = Json::decode(Json::encode($message), true)['token'] ?? null;
 
             if ($token) {
-                throw NotFound::becauseTokenNotFound($token);
+                throw NotFound::becauseTokenNotFound($token, $e->errors());
             }
 
             throw $e;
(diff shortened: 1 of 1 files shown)


jeromegamez:
This was a good catch, thank you very much! I merged this with https://github.com/kreait/firebase-php/commit/482d803637fa5e9fc303c921ba1f8d927b3d9559 - I didn't change your code, just added the tests to ensure the fixed behavior, and updated the README.

Since you were so kind to contribute, and you shared that you are stuck with 5.x (and because I checked if it was easy to do 😅), I backported your fix to the older versions and published new releases for all of them:

* https://github.com/kreait/firebase-php/releases/tag/5.26.4
* https://github.com/kreait/firebase-php/releases/tag/6.9.4
* https://github.com/kreait/firebase-php/releases/tag/7.0.1

And just like that, you made three commits instead of just one 😅 

Thanks again!

pelletiermaxime:
Wow I didn't expect that! Thank you so much, amazing work on the library!

PR #815, opened 2023-07-18
combine condition when decode service account

thinhbuzz:


jeromegamez:
Thank you for taking the time to create this PR, it's much appreciated.

However, for me this is not an optimization: yes, it saves one try/catch block, but the conditions are now more difficult to read (at least to me, and I maintain the code 😅). I usually prefer to have the code a little more verbose, but easier to parse, so I'm fine with having two `is_string()` checks here but no nesting.

PR #825, opened 2023-09-05
Add getter token in NotFound Exception

kitro:
Use case:
When using the firebase messaging service, we store the device token to send the the push notifications. sometimes we ended with unregistered token from the firebase project. the api throw an error and is catch by `NotFound` exception.
As backend service is no need to use that token that was unregistered. it will be good to know which `token` that was throw in the exception because the user it may have multiple device token stored in the backend.

### DIFF
--- a/src/Firebase/Exception/Messaging/NotFound.php
+++ b/src/Firebase/Exception/Messaging/NotFound.php
@@ -12,15 +12,13 @@ final class NotFound extends RuntimeException implements MessagingException
 {
     use HasErrors;
 
-    private string $token;
+    private ?string $token = null;
 
     /**
      * @param array<mixed> $errors
      */
     public static function becauseTokenNotFound(string $token, array $errors = []): self
     {
-        $this->token = $token;
-        
         $message = <<<MESSAGE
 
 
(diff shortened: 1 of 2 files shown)


jeromegamez:
Thanks for taking the time to create a PR! Could you please rename the `getToken()` method to `token()` and add tests to cover the method?

As a general remark: it's always nice to add a PR description with some additional insights, for example why you need this method and in which context. I know sometimes the title and changes already transport the use, but it adds a human touch 😅

kitro:
> Thanks for taking the time to create a PR! Could you please rename the `getToken()` method to `token()` and add tests to cover the method?
> 
> As a general remark: it's always nice to add a PR description with some additional insights, for example why you need this method and in which context. I know sometimes the title and changes already transport the use, but it adds a human touch 😅

Yes sure, I will add the test 👍

mlab817:
This feature would be useful in cleaning up invalid tokens. Waiting for this to be accepted.

jeromegamez:
@kitro Thank you!

MGatner:
🥳

PR #829, opened 2023-10-09
Move google/cloud-firestore from require-dev to require in composer.json

frankperez87:
## Problem
While trying to interact with the database using the facade, I encountered the `Kreait\Firebase\Exception\RuntimeException` error, indicating the `Google\Cloud\Firestore\FirestoreClient` class was not found.

## Cause
Upon investigation, I found that the `google/cloud-firestore` package was listed under `require-dev` in the `composer.json` file. This means it's not being installed in production environments, leading to the above error.

## Solution
I've moved the `google/cloud-firestore` package from `require-dev` to `require` in the `composer.json` file to ensure it's installed in all necessary environments.

## Changes
- Moved `google/cloud-firestore` to the `require` section of `composer.json`.

## Testing
I've tested the changes locally by interacting with the database using the facade, and the error no longer appears. I also ran all existing tests to ensure no regressions.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,6 +34,7 @@
         "firebase/php-jwt": "^6.3.2",
         "google/auth": "^1.24",
         "google/cloud-core": "^1.48.1",
+        "google/cloud-firestore": "^1.37.0",
         "google/cloud-storage": "^1.30.1",
         "guzzlehttp/guzzle": "^7.5",
         "kreait/firebase-tokens": "^4.2",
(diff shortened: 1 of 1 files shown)


jeromegamez:
Thanks for taking the time to create and explain the PR, it's much appreciated! 🙏🏻

[The `google/cloud-firestore` library requires the `ext-grpc` extension](https://github.com/googleapis/google-cloud-php-firestore/blob/225b40ac88b1a880df4445b83c98b656fa881c36/composer.json#L8) which not everybody has installed, and not everybody is using Firestore.

That's the reason it is in the `suggest` section instead of the `require` section. Moving the dependency would break the SDK for existing users and prevent new users from using it.

This is also already in the docs at https://firebase-php.readthedocs.io/en/stable/cloud-firestore.html

Thank you nonetheless!

MGatner:
😮‍💨

PR #835, opened 2023-10-28
Fix CS

szepeviktor:
📢 Rare event!

Only 1 type of error: `function_declaration`

Please consider running php-cs-fixer on each push. These commits affecting many files are not the nicest things in the world.

### DIFF
--- a/src/Firebase/Auth.php
+++ b/src/Firebase/Auth.php
@@ -88,7 +88,7 @@ final class Auth implements Contract\Auth
 
     public function getUsers(array $uids): array
     {
-        $uids = array_map(static fn ($uid) => Uid::fromString($uid)->value, $uids);
+        $uids = array_map(static fn($uid) => Uid::fromString($uid)->value, $uids);
 
         $users = array_fill_keys($uids, null);
 
@@ -490,7 +490,7 @@ final class Auth implements Contract\Auth
         $provider = array_values(
             array_filter(
                 array_map('strval', (array) $provider),
-                static fn (string $value) => $value !== '',
+                static fn(string $value) => $value !== '',
             ),
         );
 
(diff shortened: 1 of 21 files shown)


jeromegamez:
I usually do,  it thanks for the suggestion. That must have been a rule I forgot to migrate when I changed the rules a few weeks ago, I'll check

szepeviktor:
There is an imperfect way.
If you run php-cs-fixer in a workflow like `php-cs-fixer || true` you can get check annotations in PR diffs but still have a ✔️

image

jeromegamez:
It might be imperfect, but that's how I do it. CS is something I usually run when I push, and when I forget, I do it afterwards and (brace yourself) might even force-push it.

By the way, when I run PHP CS Fixer, it don't get the same report

CleanShot 2023-10-28 at 22 51 10@2x

jeromegamez:
I take that back, it reported them when I updated the composer dependencies :D

szepeviktor:
All problems come from humans.
That is why I've spent the last decade with inserting QA tools between humans and the application/package.

jeromegamez:
Thanks!

jeromegamez:
Oh, I know and I agree, and in my professional work-life I apply the same principles. But this started as a personal project a long time ago and still is my playground, and I like doing things by hand, and nobody has complained so far. Until I get paid for my open source work so that I can make a living, I will continue to do work on my open source projects as if they were a hobby, which they obviously are.

szepeviktor:
> Oh, I know and I agree, and in my professional work-life I apply the same principles.

Please give me something to do, where they allow me - without compensation - to clean things up from the ground up.
I pay you €1000/project.

jeromegamez:
I'm not sure how you mean this, so I don't dare to reply 😅

szepeviktor:
In the last 3 years I had ~8 hours work/week. I'd like to work in a team to make others happy.
I work with underlying technologies - the things no one has time for.

szepeviktor:
... while doing secret guerrilla monitoring I've learned that almost no team is in control of their software product. The Software has a grip on the team!!

I could tame any software that is made from my bytes (0-255).

PR #841, opened 2023-11-19
Fix regexp in lycheeignore

szepeviktor:


jeromegamez:
It worked, but I'm fine either way, thanks!

PR #850, opened 2023-12-01
Allow psr/http-message v1.0

ankurk91:
Hi,

Look like during the release of [7.7.0](https://github.com/kreait/firebase-php/compare/7.6.0...7.7.0), support for `psr/http-message` v1.0 was dropped accidently, while v1.0 can still be supported.

It would be helpful to bring back the support for `psr/http-message` v1.0 because many of legacy php packages depends on v1.x .

Thanks.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -46,7 +46,7 @@
         "psr/clock": "^1.0",
         "psr/http-client": "^1.0",
         "psr/http-factory": "^1.0",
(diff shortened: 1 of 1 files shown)


jeromegamez:
Thanks!

ankurk91:
Can you tag a minor version as well?

PR #879, opened 2024-04-11
composer-dependency-analyser: update to 1.5.0 (support function detection)

janedbal:
- Limitation with functions is finally gone
  - https://github.com/shipmonk-rnd/composer-dependency-analyser/releases/tag/1.5.0


jeromegamez:
That's great news! Thanks for the PR and your work at all! 🙌🏻

PR #890, opened 2024-05-15
Add versionNumber to RemoteConfig getTemplate

pglaum:
I want to fetch an older RemoteConfig version in my project, so I implemented the query parameter.

(The related Firebase API documentation is here: <https://firebase.google.com/docs/reference/remote-config/rest/v1/projects/getRemoteConfig>)


jeromegamez:
This is a good addition, and great to see that you followed the project's commit message and general style, thank you!

I'll merge this as-is and make some changes afterward (adding `Version` and `null` as additional argument types, as well as tests)

Thanks again!

PR #908, opened 2024-06-28
Fix Messaging for outdated environments

### LINKED ISSUE #888: Unable to send messaging after updating to 7.10.0
### Describe the bug

Running environment: PHP8.1 / "kreait/firebase-php":7.10.0

My local development environment is: windows

The online server environment is: Debian GNU/Linux 12 x86_64 (Py3.7.16)

When I was developing locally, I had no problem using 7.10.0 and could send messages normally. When my online server was updated to 7.10.0, I could not send messages. The error message was:
`URL error 0: The cURL request was retried 3 times and did not succeed. The most likely reason for the failure is that cURL was unable to rewind the body of the request and subsequent retries resulted in the same error. Turn on the debug option to see what went wrong. See https://bugs.php.net/bug.php?id=47204 for more information. (see https://curl.haxx.se/libcurl/c/libcurl-errors.html) for https://fcm.googleapis.com/v1/projects/flutter-receive-sms/messages:send`

Additionally an error page is returned.
`411 Error. POST requests require a Content-length header. That’s all we know`

### Installed packages

```json
"kreait/firebase-php":7.10.0
```

### PHP version and extensions

```shell
-
```

### Steps to reproduce the issue.

```php
    public function sendCloudMessaging(string $sendType, array $params): array|bool
    {
        Notification::fromArray([
           'title' => $params['title'],
           'body' => $params['body']
        ]);

        if (!isset($data)) {
            $params['data'] = ['jump' => 'my'];
        }
        $notification = Notification::create($params['title'], $params['body']);
        $notification = $notification->withTitle($params['title'])->withBody($params['body']);
        CloudMessage::withTarget($sendType, $params[$sendType])
            ->withNotification($notification)
            ->withData($params['data']);
            $sendType => $params[$sendType],
            'notification' => $notification,
            'data' => $params['data']
        ]);
        try {
            if (is_array($params['token'])){
                if (count($params['token']) > 500){
                    return false;
                }
                $result = $this->messaging->sendMulticast($message, $params['token']);
                return ['success' => $result->successes()->count(), 'fail' => $result->failures()->count()];
            }else{
                $result = $this->messaging->send($message);
                if (array_key_exists('name', $result)){
                    return ['success' => 1, 'fail' => 0];
                }
                return false;
            }
        } catch (MessagingException | FirebaseException $e) {
            Log::error('sendCloudMessaging error' . $e->getMessage());
            return false;
        }
    }
```

### Error message/Stack trace

```shell
-
```

### Additional information

_No response_

jeromegamez:
Fixes #888

### DIFF
--- a/src/Firebase/Messaging/RequestFactory.php
+++ b/src/Firebase/Messaging/RequestFactory.php
@@ -1,0 +1,47 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Kreait\Firebase\Messaging;
+
+use Beste\Json;
+use Psr\Http\Message\RequestFactoryInterface;
+use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\StreamFactoryInterface;
+
+/**
+ * @internal
+ */
+final class RequestFactory
+{
+    public function __construct(
+        private readonly RequestFactoryInterface $requestFactory,
+        private readonly StreamFactoryInterface $streamFactory,
+    ) {
+    }
+
+    public function createRequest(Message $message, string $projectId, bool $validateOnly): RequestInterface
+    {
+        $request = $this->requestFactory
+            ->createRequest(
+                'POST',
+                'https://fcm.googleapis.com/v1/projects/'.$projectId.'/messages:send',
+            )
+        ;
+
+        $payload = ['message' => $message];
+
+        if ($validateOnly === true) {
+            $payload['validate_only'] = true;
+        }
+
+        $body = $this->streamFactory->createStream(Json::encode($payload));
+
+        return $request
+            ->withProtocolVersion('2.0')
+            ->withBody($body)
+            ->withHeader('Content-Type', 'application/json; charset=UTF-8')
+            ->withHeader('Content-Length', (string) $body->getSize())
+        ;
+    }
+}
--- a/src/Firebase/Messaging/RequestFactory.php
+++ b/src/Firebase/Messaging/RequestFactory.php
@@ -14,10 +14,13 @@ use Psr\Http\Message\StreamFactoryInterface;
  */
 final class RequestFactory
 {
+    private bool $environmentSupportsHTTP2;
+
     public function __construct(
         private readonly RequestFactoryInterface $requestFactory,
         private readonly StreamFactoryInterface $streamFactory,
     ) {
+        $this->environmentSupportsHTTP2 = self::environmentSupportsHTTP2();
     }
 
     public function createRequest(Message $message, string $projectId, bool $validateOnly): RequestInterface
@@ -29,6 +32,10 @@ final class RequestFactory
             )
         ;
 
+        if ($this->environmentSupportsHTTP2) {
+            $request = $request->withProtocolVersion('2.0');
+        }
+
         $payload = ['message' => $message];
 
         if ($validateOnly === true) {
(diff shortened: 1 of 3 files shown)


jeromegamez:
@Daniel-Cong @david-windsock I started with doing the `'CURL_HTTP_VERSION_2*` constant check - even though we already know that this doesn't seem to be the problem, according to several comments in #888. Please test the PR anyway, so we can work towards finding the real culprit.

@rickclephas @tonysilva16 @woodongwong Since you participated in #874, I'd greatly appreciate if you could follow this PR and test it, to make sure nothing breaks for you (break = the performance gain is gone). Thank you!

david-windsock:
Thanks @jeromegamez, will try to test it on monday.

jeromegamez:
Upon further research, I stumbled upon https://github.com/microsoftgraph/msgraph-sdk-php/issues/854 and https://github.com/microsoftgraph/msgraph-sdk-php/pull/1120

I will implement this, let's hope that this is the way to check for HTTP/2 support 🤞🏻

src/Firebase/Messaging/RequestFactory.php
```diff
@@ -0,0 +1,68 @@
...
+    /**
+     * @see https://github.com/microsoftgraph/msgraph-sdk-php/issues/854
+     * @see https://github.com/microsoftgraph/msgraph-sdk-php/pull/1120
+     */
+    private static function environmentSupportsHTTP2(): bool
+    {
+        $features = curl_version()["features"] ?? null;
+
+        return
+            extension_loaded('curl')
```
rickclephas: Shouldn't we check for the loaded extension before calling `curl_version()`?

jeromegamez: Ah, you're absolutely right! Will fix!

rickclephas (commented):
I have a dependency conflict for `lcobucci/jwt` that prevents me from doing a quick test. But based on the changes I am not expecting this to break our implementation/performance 👍🏻.

src/Firebase/Messaging/RequestFactory.php
jeromegamez: Fixed!

BTW, the reason why the method is more verbose than the one in https://github.com/microsoftgraph/msgraph-sdk-php/pull/1120 is so that PHPStan doesn't complain about `curl_version()["features"]` returning `array|false`.

rickclephas:
> But based on the changes I am not expecting this to break our implementation/performance 👍🏻.

To confirm this I have tested the `environmentSupportsHTTP2` function in our project,
and it correctly returns `true` in our Google App Engine environment.

david-windsock:
> Upon further research, I stumbled upon [microsoftgraph/msgraph-sdk-php#854](https://github.com/microsoftgraph/msgraph-sdk-php/issues/854) and [microsoftgraph/msgraph-sdk-php#1120](https://github.com/microsoftgraph/msgraph-sdk-php/pull/1120)
> 
> I will implement this, let's hope that this is the way to check for HTTP/2 support 🤞🏻

We tested it on our CentOS server and worked without problem. We _faked_ the composer dependency to use this branch as the 7.9.1 version that is working for us and the notifications are sent OK, without any change on our code.

Thanks for your quick response!

jeromegamez:
That's great to hear, thank you for the feedback! 🙏🏻

I will merge this later and create a new release.

jeromegamez:
It's still curious - some have tried just commenting out the `$request = $request->withProtocolVersion('2.0');` line and it didn't seem to have worked for them. Let's hope for the best 😅

david-windsock:
Thanks @jeromegamez!

PR #960, opened 2024-11-18
Composer Dependency Analyser: update to analyse `ext-*`

janedbal:
- Updates analyser to [latest version](https://github.com/shipmonk-rnd/composer-dependency-analyser/releases/tag/1.8.0) that checks even `ext-*` dependencies
- **Removes** reported dead dependency on `ext-openssl` (which seems valid)


jeromegamez:
Looks good, thank you very much!

Issue #988, opened 2025-02-20
Notification not sent due to Error 502 from Firebase server


elia-ba:
### Describe the bug

We've just received a Sentry error report about an error generated by `firebase-php` during API communication with Firebase. Strangely enough the error reported back is an HTML page rather than a JSON error, which may point to a problem in Firebase's server but we never received this error before.

### Installed packages

```json
aws/aws-crt-php                                  1.2.7                       AWS Common Runtime for PHP
aws/aws-sdk-php                                  3.339.9                     AWS SDK for PHP - Use Amazon Web Services in your PHP project
backup-manager/backup-manager                    3.1.0                       A framework agnostic database backup manager with user-definable procedures and support for S3, Dropbox, FTP, SF...
bacon/bacon-qr-code                              2.0.8                       BaconQrCode is a QR code generator for PHP.
beberlei/doctrineextensions                      1.5.0                       A set of extensions to Doctrine 2 that add support for additional query functions available in MySQL, Oracle, Po...
beste/clock                                      3.0.0                       A collection of Clock implementations
beste/json                                       1.5.1                       A simple JSON helper to decode and encode JSON
brick/math                                       0.12.1                      Arbitrary-precision arithmetic library
brick/varexporter                                0.5.0                       A powerful alternative to var_export(), which can export closures and objects without __set_state()
cakephp/core                                     4.5.9                       CakePHP Framework Core classes
cakephp/database                                 4.5.9                       Flexible and powerful Database abstraction library with a familiar PDO-like API
cakephp/datasource                               4.5.9                       Provides connection managing and traits for Entities and Queries that can be reused for different datastores
cakephp/utility                                  4.5.9                       CakePHP Utility classes such as Inflector, String, Hash, and Security
cloud-solutions/surveylab                        dev-tlu-work-stress 0a7a5f6 SurveyLab library
composer/pcre                                    3.3.2                       PCRE wrapping library that offers type-safe preg_* replacements.
composer/semver                                  3.4.3                       Semver library that offers utilities, version constraint parsing and validation.
dasprid/enum                                     1.0.6                       PHP 7.1 enum implementation
doctrine/annotations                             2.0.2                       Docblock Annotations Parser
doctrine/cache                                   2.2.0                       PHP Doctrine Cache library is a popular cache implementation that supports many different drivers such as redis,...
doctrine/collections                             2.2.2                       PHP Doctrine Collections library that adds additional functionality on top of PHP arrays.
[...]

PR #989, opened 2025-02-25
feat: Add accept header to HTTP requests

theofidry:
Recently we had a POST to https://fcm.googleapis.com that failed (a 502). The response was more noisy than necessary because it was HTML. Ideally the response could be a simple JSON or plain text instead.

I am not sure if the Google API will respect the Accept header in this scenario, my experience with firebase is admittedly fairly limited, but it I don't think it hurts adding it?


jeromegamez:
I agree, it certainly doesn't hurt 😅. Also, interesting timing, we had the same problem a few days ago (https://github.com/kreait/firebase-php/issues/988) and I tried to mitigate it with https://github.com/kreait/firebase-php/commit/f2652f4813dc791bb820d87fd1a5e82085a1710b. Let's hope the changes help!

PR #996, opened 2025-04-02
Update SetApnsContentAvailableIfNeeded.php

khattaksd:
- data array only supports strings
- our fcm 'data' only message fails to deliver because 1 is not string

please see https://github.com/laravel-notification-channels/fcm/issues/220

### DIFF
--- a/src/Firebase/Messaging/Processor/SetApnsContentAvailableIfNeeded.php
+++ b/src/Firebase/Messaging/Processor/SetApnsContentAvailableIfNeeded.php
@@ -43,7 +43,7 @@ final class SetApnsContentAvailableIfNeeded
             return $message;
         }
 
-        $apnsConfig = $apnsConfig->withApsField('content-available', 1);
+        $apnsConfig = $apnsConfig->withApsField('content-available', '1');
 
         $payload['apns'] = $apnsConfig->toArray();
 


jeromegamez:
The APNS Config is not part of the `data` message property, could you please expand on the relation here?

khattaksd:
Agreed. It is not part of the data property. 
#### Ref: This is the log message we are seeing
```
{
  "message": {
    "data": {
      "notification_id": "2464b53d-464f-420d-88b0-aa6f9d203e0a",
      "tenant_id": "1",
      "channel_id": "dr**e",
      "android_channel_id": "dr**e",
      "action": "full_sync"
    },
    "android": {
      "priority": "high"
    },
    "token": "fhfpa4U....kQw",
    "apns": {
      "headers": {
        "apns-push-type": "background"
      },
      "payload": {
        "aps": {
          "content-available": 1
        }
      }
    }
  }
}
```
please note that the only _non-string_ value is `"{ content-available":1 }`
FCM message works if that key/value is not there. Let me know if I am misunderstanding something.
Thanks! 🙏🏾

khattaksd:
@jeromegamez you are right. this pr does not fix my problem. 😢

khattaksd:
just proved my self wrong. Testing locally by overriding the class in my project .. got this error from FCM
```
{
  "error": {
    "code": 400,
    "message": "Value for APS key [content-available] is either 0 or 1.",
    "status": "INVALID_ARGUMENT",
    "details": [
      {
        "@type": "type.googleapis.com/google.rpc.BadRequest",
        "fieldViolations": [
          {
            "field": "message.apns.payload.aps.content-available",
            "description": "Value for APS key [content-available] is either 0 or 1."
          }
        ]
      },
      {
        "@type": "type.googleapis.com/google.firebase.fcm.v1.FcmError",
        "errorCode": "INVALID_ARGUMENT"
      }
    ]
  }
}
```
I closed the PR. 
I apologize for wasting your time. Working with FCM is not fun 💢

jeromegamez:
You didn't waste my time at all, and I do agree with you 😅

What's the error that you are getting with the payload that should be working?

khattaksd:
we are sending a `data` message to apple device, FCM says 200 OK but the device does not receive the message at all. We are thinking that the `apns` property is missing some essential attributes may be. Will keep you posted.

jeromegamez:
It wouldn't be the first time that FCM had delivery issues, perhaps this is one of those cases 🤞🏻

Another thing that might help: could you try explicitly setting/duplicating the data property in the `apns` configuration, just to see if that leads to a successful delivery?

khattaksd:
### (Mis)Adventures in iOS background push message using FCM
`Data Message` aka `Background Message`

- iOS background message is not the same as Android background message - both in semantics and implementation
- While Android handles data message seamlessly and even encourages it, iOS throttles and discourages it
- We resorted to sending push message w/o `notification` attribute in the json so that `android` can interpret it as a data message and for `apns` we made it look like an alert like by doing so:
```
])->custom([
    'android' => ['priority' => 'high'],
    'apns' => [
        'payload' => [
            'aps' => [
                'sound' => 'default',
                'content-available' => 1,
            ],
        ],
        'headers' => [
            'apns-push-type' => 'alert',
            'apns-priority' => '10',
        ],
    ],
]);
```

PR #1003, opened 2025-06-05
Enable getting a user by their federated IdP UID

jeromegamez:
This is based on and replaces #1000 with the following changes/additions:

- Changed tests so they run within the emulator test suite and not in the integration test suite
  - This allows for a happy path and an unhappy path test
  - Instead of using a dummy IdP, the real™/emulated `phone` IdP can be used
- Updated documentation and changelog

Thank you @mkilmanas for the great work and collaboration in #1000, the commit of this PR is attributed to us both!

:octocat:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,34 @@ Please read about the future of the Firebase Admin PHP SDK on the
 
 ## [Unreleased]
 
+### Added
+
+* You can now get a user by their federated identity provider (e.g. Google, Facebook, etc.) UID with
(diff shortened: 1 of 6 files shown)


mkilmanas:
Hey @jeromegamez - thanks for pushing this though the finish line. However, I see that this PR was merged 1 day after the latest release was tagged. Are there any plans when the next release might get tagged? Still waiting for this functionality to become publicly available 😄

jeromegamez:
Shoot, I thought I did! I will create a new release later today!

jeromegamez:
Have I mentioned that my days have 48h and that's why you _think_ I'm late but I'm actually not! 😅 

Anyway, at last, the release is here: https://github.com/kreait/firebase-php/releases/tag/7.20.0

Thanks again!

PR #1007, opened 2025-06-19
Add conditionable trait to be used in classes

fului:
## What does it do
Add a conditionable trait to be able to do the following.

```php
$config = ApnsConfig::new()
    ->when($deeplink !== null, fn(ApnsConfig $config) => $config->withApsField('custom_data', ['deeplink' => $deeplink]));

$message = CloudMessage::new()
    ->when($deeplink !== null, fn (CloudMessage $message) => $message->withData(['deeplink' => $deeplink]));
```

I know that this was already achievable with an if statement like this

```php
$message = CloudMessage::new();

if ($deeplink !== null) {
    $message = $message->withData(['deeplink' => $deeplink']);
}
```

So this PR is only if you feel like this would benefit the project.

## What has changed
Currently I have only added this to the `CloudMessage`, `ApnsConfig` and `AndroidConfig` classes and created tests for these classes.

### DIFF
--- a/src/Firebase/Concerns/Conditionable.php
+++ b/src/Firebase/Concerns/Conditionable.php
@@ -1,0 +1,46 @@
+<?php
+
+namespace Kreait\Firebase\Concerns;
+
+trait Conditionable
+{
+    /**
+     * Conditionally apply a callback if the given condition is true.
+     *
+     * @param bool $condition The condition to evaluate.
+     * @param callable $callback The callback to execute if condition is true.
+     * @param callable|null $default The fallback callback if condition is false.
+     * @return $this|mixed
+     */
+    public function when(
+        bool $condition,
+        callable $callback,
+        ?callable $default = null,
+    ): mixed {
+        if ($condition) {
+            return $callback($this);
+        }
+
+        if ($default !== null) {
+            return $default($this);
+        }
+
+        return $this;
+    }
+
+    /**
+     * Conditionally apply a callback if the given condition is false.
+     *
+     * @param bool $condition The condition to evaluate.
+     * @param callable $callback The callback to execute if condition is false.
+     * @param callable|null $default The fallback callback if condition is true.
+     * @return $this|mixed
+     */
+    public function unless(
+        bool $condition,
+        callable $callback,
+        ?callable $default = null,
+    ): mixed {
+        return $this->when(! $condition, $callback, $default);
+    }
+}
--- a/src/Firebase/Messaging/AndroidConfig.php
+++ b/src/Firebase/Messaging/AndroidConfig.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Kreait\Firebase\Messaging;
 
 use JsonSerializable;
+use Kreait\Firebase\Concerns\Conditionable;
 use Kreait\Firebase\Exception\Messaging\InvalidArgument;
 
 use function array_filter;
@@ -81,6 +82,8 @@ use function sprintf;
  */
 final class AndroidConfig implements JsonSerializable
 {
+    use Conditionable;
+
     private const MESSAGE_PRIORITY_NORMAL = 'normal';
 
     private const MESSAGE_PRIORITY_HIGH = 'high';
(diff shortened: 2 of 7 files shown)


jeromegamez:
Thank you for taking the time to think about improving the SDK and for creating a PR - with tests, no less! That's relatively rare in this repo 😅

I don't feel comfortable including this kind of *syntactic sugar* (for lack of a better term) directly in the SDK - at least not at this time. It adds more code to maintain and sits a bit outside the core domain of the SDK.

Laravel has this particular helper, but - depending on who you ask - it's debatable whether it's actually "nicer" to work with. If I accepted this PR, someone else might come along later and want to add `Macroable`, pointing back to this as precedent when I decline 😅

Yes, wrapping a dozen conditions in `if` statements takes up space - but that can be addressed with short ternaries, which can end up even more concise than `->when()`:

```php
// Using `when()`
$message = $message->when(
    $deeplink !== null,
    fn (CloudMessage $message) => $message->withData(['deeplink' => $deeplink])
);

// vs ternary
$message = $deeplink !== null ? $message->withData(['deeplink' => $deeplink]) : $message;
```

`->when()` enables chaining, but short ternaries don't look too bad either:

```php
// Using `when()` chaining
$message = CloudMessage::new()
    ->when($cond1, fn (CloudMessage $m) => ...)
    ->when($cond2, fn (CloudMessage $m) => ...)
    ->when($cond3, fn (CloudMessage $m) => ...);

// vs ternary
$message = CloudMessage::new();
$message = $cond1 ? ... : $message;
$message = $cond2 ? ... : $message;
$message = $cond3 ? ... : $message;
```

Short ternary syntax also places fallback/default behavior right at the callsite - with `Conditionable`, that ergonomics advantage diminishes:

```php
$defaultLink = '...';

$message = CloudMessage::new()
    ->when(
        $deeplink !== null,
        fn (CloudMessage $m) => $m->withData(['deeplink' => $deeplink]),
        fn (CloudMessage $m) => $m->withData(['deeplink' => $defaultLink])
    );

// vs
$message = $message->withData([
    'deeplink' => $deeplink !== null ? $deeplink : $defaultLink,
]);
```

And if the callback is more involved than a simple method call, one might argue that a good old `if` block is easier to follow:

```php
$defaultLink = 'default link';
$deeplinkWasSet = null;

$message = $message->when(
    $deeplink !== null,
    function (CloudMessage $message) use (&$deeplinkWasSet) {
        $deeplinkWasSet = true;
        return $message->withData(['deeplink' => $deeplink]);
    }
);

// vs
if ($deeplink !== null) {
    $message = $message->withData(['deeplink' => $deeplink]);
    $deeplinkWasSet = true;
}
```

Also, a note on `withData()`: it can't be called multiple times - each call overwrites the previous data payload. So if you had multiple conditions controlling what goes into `data`, you'd need to do something like this outside of `when()` anyway:

```php
$message = $cond1 ? $message->withData([...]) : $message;
$message = $cond2 ? $message->withData([...]) : $message; // Overwrites previous data

// Better:
$data = array_filter([
    'deeplink' => $deeplink !== null ? $deeplink : null,
    'text' => $text !== '' ? $text : null,
    'number' => $number > 0 ? $number : null,
]);

$message = $message->withData($data);
```

Testing, debugging, and static analysis are another concern - with `when()`, you'd have to step into another layer of indirection (conditionable + closures), instead of just dropping a breakpoint inside a simple `if`/`else`.

My examples are deliberately specific - and you might not agree with how accurate or "real-world" they are - but that's sort of the point. `if/else` isn't pretty, but it's unambiguous and leaves full control in the developer's hands.

Just to be clear: I didn't write this wall of text to "destroy" your PR - it absolutely has merit. `when()` exists in Laravel/Eloquent, and it *does* make things look nicer in many cases. That's why I felt the need to explain why *I* don't think it belongs in the SDK - at least not directly. (Also: to have something to refer back to if a similar PR or feature request comes up later.)

All that being said: A genuine thank-you for taking the time to create the PR - even if this one doesn't make it in! 🌺

fului:
Hi Jérôme,
thank you for taking the time to write this thorough explanation of why you had to reject this PR. 

Your reasons make perfect sense and I will keep this in mind, if I come across another PR for your project.

Thank you for maintaining this project, for your commitment and communication, I really appreciate you spending the time on your explanations in issues and PR's. :) 

Best regards,
Olaf

jeromegamez:
Thank you for your understanding and the kind words! 🫶🏻

PR #1009, opened 2025-07-13
Introduce Valinor for type-safe object mapping

jeromegamez:
Introduce [Valinor](https://valinor.cuyz.io/) for type-safe object mapping. The first application is mapping a given service account file, JSON, or array to the newly added internal `ServiceAccount` class, with more to follow in future releases.

:octocat:


src/Firebase/ServiceAccount.php
```diff
@@ -0,0 +1,54 @@
...
+
+use SensitiveParameter;
+
+/**
+ * @internal
+ */
+final class ServiceAccount
+{
+    /**
+     * @param non-empty-string $type
```
romm: Not a big deal, but just so you know: you can use PHP docs on promoted properties. IMHO the closer the property and the type declaration are, the better.

```php
    public function __construct(
        /** @var non-empty-string */
        public string $type,
        #[SensitiveParameter]
        /** @var non-empty-string */
        public string $projectId,
        …
    } ()
```

src/Firebase/Valinor/Mapper.php
```diff
@@ -0,0 +1,71 @@
...
+        if ($cache !== null) {
+            $builder = $builder->withCache($this->cache);
+        }
+
+        $this->mapperBuilder = $builder;
+    }
+
+    public function withConverter(callable $converter): self
+    {
+        $mapperBuilder = $this->mapperBuilder->registerConverter($converter); // @phpstan-ignore-line argument.type
```
romm, line 32: I guess you get a PHPStan message about `pure-callable`? Instead of ignoring the error, you could type-hint the `$converter` parameter of this method with `@param pure-callable $converter`. And then, marking the converters passed to this method with the `@pure` annotation.

src/Firebase/Valinor/Normalizer.php
```diff
@@ -0,0 +1,82 @@
...
+        if ($cache !== null) {
+            $builder = $builder->withCache($this->cache);
+        }
+
+        $this->normalizerBuilder = $builder;
+    }
+
+    public function withTransformer(callable $transformer): self
+    {
+        $builder = $this->normalizerBuilder->registerTransformer($transformer); // @phpstan-ignore-line argument.type
```
romm, line 43: Same message for converters in the mapper regarding purity 😊

romm:
Some minor comments, but overall LGTM! 👍

romm:
Just saw that the review was not submitted 🤦

src/Firebase/ServiceAccount.php
jeromegamez: Much better, thanks a lot!

src/Firebase/Valinor/Mapper.php
jeromegamez: I remember I tried that but then PHPStan complained that `strtolower()` or one of the other functions were potentially impure... I will try again!

jeromegamez:
Great feedback, thanks so much! I'll make sure to implement your suggestions once I get back at my desk!

src/Firebase/Valinor/Mapper.php
jeromegamez: `preg_replace` as well, there's an open issue about it at https://github.com/phpstan/phpstan/issues/11884

romm: Yay, that's the issue with purity, it's not completely reliable yet 🙁 

Also, see this chapter in doc: https://valinor.cuyz.io/latest/other/static-analysis/#about-mapper-and-normalizer-purity

romm:
Hi @jeromegamez just wanted to make sure if everything is fine on your side, or you need some help regarding the integration of the mapper? Have a good day!

jeromegamez:
Everything is fine, thanks a lot for asking! I'm just short on time and contemplating if I should leave it with this to introduce it in the first place or if I should wait until I have migrated all the other value objects first - but mostly short on time 😅

romm:
Alright, no worries!

I'd recommend releasing it with just a few DTO in the first place. This way, if anything goes badly for end-users (we never know) it will be easier to fix the issues. But your decision, of course. 😊

jeromegamez:
YOU'RE JUST HERE FOR THE DOWNLOADS!!! 😂

romm:
I'd be lying if I said I'm not 😁! I'm mostly interested in seeing if this works fine with a big user base, though. 😛

But I'd have given the same advice anyway! 😅

jeromegamez:
@romm What I meant to ask - would something like https://github.com/kreait/firebase-php/blob/f410f49f2209ccfb105d02cc39041d594bf89dbe/src/Firebase/Valinor/Source.php make sense to integrate into Valinor by default for the JSON Source?

romm:
> @romm What I meant to ask - would something like [`f410f49`/src/Firebase/Valinor/Source.php](https://github.com/kreait/firebase-php/blob/f410f49f2209ccfb105d02cc39041d594bf89dbe/src/Firebase/Valinor/Source.php) make sense to integrate into Valinor by default for the JSON Source?

I see the value it can bring, however I'm more of an "explicit coding" type of guy, most of the time, so I'd prefer users to use existing `Source::*` methods.

Also, JSON values can start with `[`, if you also want to support JSON lists in your own `Source::parse()` method. 😊

jeromegamez:
Oh, good point, I will add this, thank you!

jeromegamez:
Released with [7.21.0](https://github.com/kreait/firebase-php/releases/tag/7.21.0) 🤗

romm:
🚀 

I should have a bit more spare time in the upcoming days, do not hesitate to ping me if any issue related to the mapping/normalization occurs, I could probably help!

Also, it just occurred to me (missed it during the review) that no cache is used: do you plan to use it at some point? Do you already use cache in other parts of the library?

Anyway, thanks for keeping me updated!

jeromegamez:
That was on purpose - I think we discussed this on Mastodon, but I don't remember (if there was) a conclusion... 😅

I wanted to be sure that this wouldn't break production for projects using read-only filesystems, for example on AWS ECS (I have first-hand experience there) or cloud functions (with Bref).

Concerning cache - the SDK comes with a PSR-16 Memory Cache only, but others can be configured with the factory, including the mapper and normalizer cache - I did forget to update the docs, though, I think 😬

romm:
Indeed, I forgot about that, it makes sense. People who need better performance can always inject the cache themselves I guess. Updating the docs could be nice indeed. 😊

src/Firebase/Valinor/Source.php
```diff
@@ -0,0 +1,52 @@
...
+        if (str_starts_with((string) $value, '{')) {
+            try {
+                return new self(BaseSource::json($value));
+            } catch (Throwable $e) {
+                throw new InvalidArgumentException(message: $e->getMessage(), previous: $e);
+            }
+        }
+
+        try {
+            return new self(BaseSource::file(new SplFileObject($value)));
```
stephank, line 42: Just want to highlight that this was a small breaking change for us, because files are now expected to have a `.json`, `.yaml` or `.yml` extension, and this wasn't required before. Maybe an edge case, because of the way we provision config in our projects. We ended up simply reading the file ourselves before passing it to the lib.

jeromegamez: Oh, shoot, I'm sorry! How do you provision the config? Perhaps I can adapt the SDK to other file types and formats.

Update: Is it just™ the file extension?

jeromegamez: I'm addressing this in https://github.com/kreait/firebase-php/pull/1013 - I hope this will restore the previous behavior for you! 🤞🏻

PR #1028, opened 2025-10-12
Fix Valinor converter signature to match new rule

romm:
Reference: https://github.com/CuyZ/Valinor/releases/tag/2.2.1

Hi @jeromegamez, I'm submitting this just after the [`cuyz/valinor` 2.2.1 release](https://github.com/CuyZ/Valinor/releases/tag/2.2.1) which brings a bugfix that can lead to issues for your library users. As this represents a lot of people, I urged myself to write the fix before people come to complain about it.

I'm sorry about this, that's the first time I'm kind of forced to do a BC break inside a non-major, although I believe it makes sense (as explained in the release notes linked above). I recommend releasing a new version of `kreait/firebase-php` ASAP with this commit.

Note that this commit also marks `cuyz/valinor` versions before 2.2.1 as conflicting. This should prevent a lot of applications having issues that upgrade either this package or `cuyz/valinor` without upgrading the other one.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,7 +31,7 @@
         "beste/clock": "^3.0",
         "beste/in-memory-cache": "^1.3.1",
         "beste/json": "^1.5.1",
-        "cuyz/valinor": "^2.0",
+        "cuyz/valinor": "^2.2.1",
         "fig/http-message-util": "^1.1.5",
         "firebase/php-jwt": "^6.10.2",
         "google/auth": "^v1.45",
--- a/src/Firebase/Valinor/Converter/SnakeCaseToCamelCaseConverter.php
+++ b/src/Firebase/Valinor/Converter/SnakeCaseToCamelCaseConverter.php
@@ -13,6 +13,12 @@ use Traversable;
  */
 final class SnakeCaseToCamelCaseConverter
 {
+    /**
+     * @template T of object
+     * @param array<mixed> $value
+     * @param callable(array<mixed>): T $next
+     * @return T
+     */
     public function __invoke(mixed $values, callable $next): object
     {
         if ($values instanceof Traversable) {


jeromegamez:
Hey Romain! Thank you for the bugfix and taking the time to make this PR! Could we also just™ update the `require` section instead of adding the `conflict` section?

```diff
- "cuyz/valinor": "^2.0",
+ "cuyz/valinor": "^2.2.1",
```

PS: Don't mind the PHPStan errors, there an annoying, unrelated problem.
PPS: Don't mind the integration test failure, that's just™ a brittle test 😅

romm:
Not sure what the PHPStan issue is about, also I don't know why but I can't run integration tests on my local installation:

```
$ composer test-integration
PHPUnit 10.5.58 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.3.24
Configuration: /Users/romm/PhpstormProjects/firebase-php/phpunit.xml.dist

No tests executed!
```

romm:
> Hey Romain! Thank you for the bugfix and taking the time to make this PR! Could we also just™ update the `require` section instead of adding the `conflict` section?
> 
> ```diff
> - "cuyz/valinor": "^2.0",
> + "cuyz/valinor": "^2.2.1",
> ```

Actually yes, this should be enough 😅 It has been a long day (and night) of coding haha

jeromegamez:
Thanks a lot! I'll prepare a new release in an instant! 💪🏻

romm:
That's the least I could do!

I hope it wont affect end-users of your library. Please do not hesitate to ping me if anything Valinor-related happens. 😊

jeromegamez:
I hope so, too, but I'm confident that all library users always use the latest PHP and library versions, so they should be fine! 🤡 

Thanks again! 🫶🏻

romm:
My pleasure! Have a good night 😊

jeromegamez:
@romm I have a curious issue - I noticed that `@param array<mixed> $value` doesn't match the method signature (which uses `$values`). When I "fixed" it (04fa5b204ac9e294bbdf93e3bcf80c4b44841260), the tests failed because the input array couldn't be matched (my bad for not testing this locally or inspecting the related GitHub actions).

When I step-debugged it, `$values` turned out to be an instance of [Kreait\Firebase\Valinor\Source](https://github.com/kreait/firebase-php/blob/04fa5b204ac9e294bbdf93e3bcf80c4b44841260/src/Firebase/Valinor/Source.php), so accepting it (0814061a0b3654981029f8192b233f4f29c8cdfe) made it work again.

Not sure if this is the correct way to approach this - I introduced my custom `Source` to allow passing in JSON strings, perhaps there is a better way? I'm not asking you to debug this yourself, since running the tests is not straight forward in the SDK's current state 😅, but perhaps you have an idea just by looking at it…

romm:
Hi @jeromegamez, funny bug haha. What happened was:

1. Because there was a mistake in the docblock for the parameter name, the type `array<mixed>` was not assigned and the native type `mixed` was used instead
2. The type `mixed` accepts anything, so no matter what the source value was, it would always match the converter's parameter
3. Changing the name to `$values` now assigns the type `array<mixed>`, but because the source value is an instance of `Kreait\Firebase\Valinor\Source` it doesn't match the type so the converter is skipped

To make it work, I believe that changing the type to `iterable<mixed>` (the native type must **also** be changed to `iterable`) should be enough.

I'll also have to update Valinor's documentation to match this change, because you probably wont be the last person to report this issue 😅

jeromegamez:
Thanks a lot for the `iterable` hint - much better now: https://github.com/kreait/firebase-php/commit/cab1937263767843b6c41fae3b93bfcb9bb566c4)

PR #1051, opened 2025-12-22
update Firestore namespace due to Google updated it recently

lieshanhe:
update Firestore namespace due to Google updated it recently

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,5 @@
 *.log
-
+/.idea
 /.build
 /vendor
 /docs/_build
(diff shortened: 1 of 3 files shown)


jeromegamez:
Could you show me where the namespace has changed and where/when this has been announced?

https://github.com/googleapis/google-cloud-php/blob/main/Firestore/src/FirestoreClient.php

lieshanhe:
Oops, I didn’t realize that I needed to install the google/cloud-storage package. I ran into an exception saying that Google\Cloud\Firestore\FirestoreClient does not exist.I’ll close the request for now. Thanks for taking a look.

PR #1111, opened 2026-07-22
Add support for Guzzle 8

GrahamCampbell:
This adds Guzzle 8 support by widening the `guzzlehttp/guzzle`, `guzzlehttp/psr7` and `guzzlehttp/promises` constraints to allow their new majors alongside the current ones. Guzzle 8 removed `getResponse()` from the base `RequestException`, where it now lives only on the `ResponseException` subclasses, and reclassified connection failures under `Psr\Http\Client\NetworkExceptionInterface`. The five API exception converters therefore guard the response lookup so it resolves on both majors, and the Remote Config converter catches the PSR-18 network interface for connection errors. The behaviour is unchanged on Guzzle 7.

Composer can only resolve Guzzle 8 here once this project's own dependencies allow it, so this PR depends on releases from beste/firebase-tokens-php#73 (`kreait/firebase-tokens`), googleapis/google-auth-library-php#677 (`google/auth`) and googleapis/google-cloud-php#9377 (`google/cloud-storage`). Until those ship, Composer keeps resolving Guzzle 7 and this branch stays fully compatible with it.


jeromegamez:
PR 1111, nice! Thank you for your work - I'm keeping track of the PRs in the other repos and will enable the workflows once they're all ready.

~I believe https://github.com/googleapis/gax-php needs the same treatment, too.~ My mistake, it's been incorporated into the cloud core project since I last looked 😅

jeromegamez:
Thanks again, much appreciated!

PR #1115, opened 2026-08-04
Update firebase/php-jwt version constraint

aronduby:
Updated firebase/php-jwt version constraint to only allow 7.0.2 or higher. This addresses the security advisory on versions less than 7 

https://github.com/advisories/GHSA-2x45-7fc3-mxwq


jeromegamez:
Thanks for the PR! The CVE is rated as low severity and has since been [disputed](https://github.com/googleapis/php-jwt/issues/620), primarily because choosing an appropriate key length is considered the application’s responsibility rather than the library’s. I generally agree with that assessment.

Nevertheless, after reviewing the library’s [most-downloaded dependents](https://packagist.org/packages/firebase/php-jwt/dependents?order_by=downloads), it appears that most already support version 7.x. I’m therefore comfortable updating our constraint to ^7.0.2.

Thanks again, much appreciated! 🙏🏻

PR #1138, opened 2026-09-18
Messaging: Support the `fid` message target

achretien:
The FCM API [deprecated](https://firebase.google.com/docs/reference/fcm/rest/v1/projects.messages) a message's `token` field in favor of `fid`, which targets a [Firebase Installation ID](https://firebase.google.com/docs/cloud-messaging/android/get-started#access-firebase-installation-id). `token` stays fully supported until Google decommissions it and currently also accepts FIDs, so this is additive — no action required for existing code.

* `CloudMessage::withFid()`, plus `fid` support in `CloudMessage::fromArray()`. Like other targets, a FID replaces a previously set one.
* New `FirebaseInstallationId` and `FirebaseInstallationIds` value objects.
* New `MessageTarget::FID`.
* `Messaging::sendMulticast()` accepts `FirebaseInstallationId(s)`. Strings and string arrays are still registration tokens.
* `MulticastSendReport::validFids()` and `unknownFids()`.

Unit tests, PHPStan, Rector and php-cs-fixer pass. Integration tests need a `TEST_FIREBASE_INSTALLATION_IDS` env var and weren't run locally.


src/Messaging.php
```diff
@@ -280,21 +291,23 @@ private function messageHasTarget(Message $message): bool
         $check = Json::decode(Json::encode($message), true);
 
         return array_key_exists(MessageTarget::CONDITION, $check)
+            || array_key_exists(MessageTarget::FID, $check)
             || array_key_exists(MessageTarget::TOKEN, $check)
             || array_key_exists(MessageTarget::TOPIC, $check);
     }
 
-    private function withChangedTarget(Message $message, string $value): RawMessageFromArray
+    private function withChangedTarget(Message $message, string $value, string $type = MessageTarget::TOKEN): RawMessageFromArray
```
jeromegamez: Please change this to `?string $type = null` and then add a `$type ??= MessageTarget::TOKEN`

jeromegamez:
Thank you for taking the time and for adding FID support, I really appreciate it! 

However, I don't think it can be merged in its current form. 

* Widening the `sendMulticast()` method breaks implementers of the interface, renaming the argument breaks usages of named arguments.
* Adding a new possible `MessageTarget::type()` and changing the public `MessageTarget::TYPES` constant can break consumers, e.g. when doing a `match ($report->target()->type()) {}` without a default case. It's not a hard BC break, but still a compatibility risk
* `sendMulticast()` also accepts `array|string` - this makes it impossible to distinguish registration tokens from FIDs

One possible compatible approach would be a interface like I used it for [App Check replay protection](https://github.com/beste/firebase-php/commit/91cff5a497642177f2124d007b0977cdd3d4d915), for example:

```php
interface MessagingWithFidMulticast
{
    public function sendMulticastToFids(...): MulticastSendReport;
}
```

[Messaging](https://github.com/beste/firebase-php/blob/8.x/src/Messaging.php) could implement both contracts while `Contract\Messaging` remains unchanged. In the next major release we could keep but deprecate `sendMulticast()` and introduce another `sendMulticastToRegistrationTokens()`. A migration interface now could be:

```php
interface MessagingWithMulticast
{
	public function sendMulticastToRegistrationTokens(...): MulticastSendReport;
    public function sendMulticastToFids(...): MulticastSendReport;
}
```

and `sendMulticast() would then defer to `sendMulticastToRegistrationTokens`.

While researching the changes, I noticed that FCM's deprecated [token](https://firebase.google.com/docs/reference/fcm/rest/v1/projects.messages#Message.FIELDS.token)[ field already accepts FIDs](https://firebase.google.com/docs/reference/fcm/rest/v1/projects.messages#Message.FIELDS.token), so we can use FIDs in the `token` field (and `RegistrationToken` class) during the transition, even though the naming becomes inaccurate. This would require no changes for sending, but we can't know if the current app instance and topic subscription handling will continue to work.

The app-instance and topic-management code still uses the legacy Instance ID API. The replacement [topic-subscription API](https://firebase.google.com/docs/reference/fcm/fcmtopics/rest/v1/projects.registrations.topicSubscriptions#resource:-topicsubscription) accepts registration tokens _and_ FIDs, but does not replace all data returned by `getAppInstance()`. That's nothing this PR needs to solve though.

Let me know what you think!

achretien:
Thanks, all three points were fair. Reworked and force-pushed.

`sendMulticast()` is back to its exact old signature and param name, `Contract\Messaging` untouched. It just defers to `sendMulticastToRegistrationTokens()` now.

The new methods live on a `MessagingWithMulticast` contract like you suggested, same shape as `AppCheckWithReplayProtection`.

I also added a `validateFids()` there, since otherwise there's no FID equivalent of `validateRegistrationTokens()`. Makes the interface name a bit off, so rename or split it if you'd rather.

On `TYPES`: it's worse than a compatibility risk, I ran roave-bc-check and the version you reviewed had three breaks, including that constant. So `bc-check` would have gone red.

`TYPES` now keeps its old value and there's an `ALL_TYPES` next to it for the internal single-target check. Branch is at zero breaks now.

That only leaves the `match` on `type()`. 

I kept `MessageTarget::FID` since you can only ever get a `fid` target if you called `withFid()` or `sendMulticastToFids()` yourself, so nobody gets surprised by it.

But your `token` field idea works fine too (`RegistrationToken` doesn't validate anything, so FIDs already go through it today) and only costs the `validFids()` / `unknownFids()` helpers. As you said for now the `token` field accept `fid` during transition. 

One thing I'm unsure about: `messageTargetWasInvalid()` greps the error message for `token`, which won't match anything about a `fid` field, so `invalidFids()` would always come back empty. I widened it to `(token|fid)`.

src/Messaging/MulticastSendReport.php
jeromegamez: In `unknownFids()` you added

```php
->filter(static fn(SendReport $report): bool => $report->target()->type() === MessageTarget::FID)
```

but here you didn't add

```php
->filter(static fn(SendReport $report): bool => $report->target()->type() === MessageTarget::TOKEN)

File: src/Messaging/MulticastSendReport.php [hunk 2]
```diff
@@ -108,6 +108,45 @@ public function invalidTokens(): array
         ;
```
jeromegamez: In `invalidFids()` you added

```php
->filter(static fn(SendReport $report): bool => $report->target()->type() === MessageTarget::FID)
```

but here you didn't add

```php
->filter(static fn(SendReport $report): bool => $report->target()->type() === MessageTarget::TOKEN)

File: src/Messaging/MessageTarget.php [hunk 3]
```diff
@@ -21,10 +23,22 @@
...
     public const array TYPES = [
         self::CONDITION, self::TOKEN, self::TOPIC, self::UNKNOWN,
     ];
 
+    /**
+     * All target types, including self::FID.
+     */
+    public const array ALL_TYPES = [
+        self::CONDITION, self::FID, self::TOKEN, self::TOPIC, self::UNKNOWN,
+    ];
```
jeromegamez:
```suggestion
	private const array ALL_TYPES = [
        self::CONDITION, self::FID, self::TOKEN, self::TOPIC, self::UNKNOWN,
    ];
```

`TYPES` is public to preserve backward compatibility, but it's only used for internal checks, so I think this new constant should be a private.

The comment is superfluous - after this PR is merged, there's only `ALL_TYPES`, and they include all types, no need to mention `self::FID` explicitly

src/Messaging/MulticastSendReport.php
```diff
@@ -108,6 +108,45 @@ public function invalidTokens(): array
...
     }
 
+    /**
+     * @return list<non-empty-string>
+     */
+    public function validFids(): array
+    {
+        return $this->successes()
+            ->filter(static fn(SendReport $report): bool => $report->target()->type() === MessageTarget::FID)
```
jeromegamez: You filter by target type `MessageTarget::FID` here, but didn't add the a filter for `MessageTarget::TOKEN` in the `validTokens()` method, so, currently, that method includes both tokens and FIDs.

src/Messaging/MulticastSendReport.php
```diff
@@ -108,6 +108,45 @@ public function invalidTokens(): array
...
+
+    /**
+     * Returns all provided Firebase Installation IDs that were not reachable.
+     *
+     * @return list<non-empty-string>
+     */
+    public function unknownFids(): array
+    {
+        return $this
+            ->filter(static fn(SendReport $report): bool => $report->target()->type() === MessageTarget::FID)
```
jeromegamez: You filter by target type `MessageTarget::FID` here, but didn't add the a filter for `MessageTarget::TOKEN` in the `unknownTokens()` method, so, currently, that method includes both tokens and FIDs.

src/Messaging/MulticastSendReport.php
```diff
@@ -108,6 +108,45 @@ public function invalidTokens(): array
...
+
+    /**
+     * Returns all provided Firebase Installation IDs that were invalid.
+     *
+     * @return list<non-empty-string>
+     */
+    public function invalidFids(): array
+    {
+        return $this
+            ->filter(static fn(SendReport $report): bool => $report->target()->type() === MessageTarget::FID)
```
jeromegamez: You filter by target type `MessageTarget::FID` here, but didn't add the a filter for `MessageTarget::TOKEN` in the `invalidTokens()` method, so, currently, that method includes both tokens and FIDs.

tests/IntegrationTestCase.php
```diff
@@ -136,4 +155,23 @@ private static function registrationTokensFromEnvironment(): array
 
         return array_values($tokens);
     }
+
+    /**
+     * @return list<non-empty-string>
+     */
+    private static function firebaseInstallationIdsFromEnvironment(): array
+    {
+        $value = Util::getenv('TEST_FIREBASE_INSTALLATION_IDS');
```
jeromegamez: You read from `TEST_FIREBASE_INSTALLATION_IDS`, but it's not set anywhere. Could you please add the variable to `tests/.env.dist`, `CONTRIBUTING.md`, and `.github/workflows/secure-tests.yml`? The secure tests don't run on external PRs, but this way I could obtain a FID for myself, run the tests locally and add it to the CI secrets.

jeromegamez (changes requested):
Thank you for the updates, I added some comments and change requests.

When you have modified the methods in the `MulticastSendReport`, could you please add unit tests that prove that the filtering works with the different tokens? The `MulticastSendReportTest` would be a good location.

Thank you!

